PR #69, opened 2026-05-17
Backup and Restore internal database and settings.

### LINKED ISSUE #55: Scan folder
After a uninstall and reinstall, it doesn't seem like its possible to point the app to which folder has all tracks downloaded. because you can only manually import one song at a time and it doesn't include artwork.

winstriker:
Backup and restore system, allowing users to export their entire database and preferences into a single ZIP archive, ensuring data portability for reinstalling the application or switching phones.

When restoring a backup the user will be prompted to select the song library folder so that permissions will be granted again.

It is impossible to export youtube and spotify cookies as they're encrypted with a device-bound master key, so after restoring a backup the user has to log in again to the services.

- Implemented a checkpoint-then-zip approach. It forces a `wal_checkpoint(FULL)` to ensure the main `.db` file is fully synchronized before copying.
- Bundles the database file and all files within the `datastore/` directory into a ZIP.
- During import, it safely closes the database, replaces files, and cleans up stale `-wal` and `-shm` files to prevent corruption.
-  Added Backup section in settings page

it also closes #55 

I think ready to merge @rawnaldclark, lmk!

### DIFF
--- a/core/data/src/main/kotlin/com/stash/core/data/db/DatabaseBackupManager.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/db/DatabaseBackupManager.kt
@@ -1,0 +1,152 @@
+package com.stash.core.data.db
+
+import android.content.Context
+import android.net.Uri
+import androidx.sqlite.db.SimpleSQLiteQuery
+import dagger.hilt.android.qualifiers.ApplicationContext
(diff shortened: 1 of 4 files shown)


winstriker:
sorry made a mess with branches, gonna reopen

PR #70, opened 2026-05-17
 Backup and Restore internal database and settings

### LINKED ISSUE #55: Scan folder

winstriker:
Backup and restore system, allowing users to export their entire database and preferences into a single ZIP archive, ensuring data portability for reinstalling the application or switching phones.

When restoring a backup the user will be prompted to select the song library folder so that permissions will be granted again.

It is impossible to export youtube and spotify cookies as they're encrypted with a device-bound master key, so after restoring a backup the user has to log in again to the services.

- Implemented a checkpoint-then-zip approach. It forces a `wal_checkpoint(FULL)` to ensure the main `.db` file is fully synchronized before copying.
- Bundles the database file and all files within the `datastore/` directory into a ZIP.
- During import, it safely closes the database, replaces files, and cleans up stale `-wal` and `-shm` files to prevent corruption.
-  Added Backup section in settings page

it also closes #55 

I think ready to merge @rawnaldclark, lmk!

### DIFF
--- a/core/data/src/main/kotlin/com/stash/core/data/db/DatabaseBackupManager.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/db/DatabaseBackupManager.kt
@@ -1,0 +1,152 @@
+package com.stash.core.data.db
+
+import android.content.Context
+import android.net.Uri
+import androidx.sqlite.db.SimpleSQLiteQuery
+import dagger.hilt.android.qualifiers.ApplicationContext
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import java.io.File
+import java.io.FileOutputStream
+import java.io.FileInputStream
+import java.util.zip.ZipEntry
+import java.util.zip.ZipInputStream
+import java.util.zip.ZipOutputStream
+import javax.inject.Inject
+import javax.inject.Singleton
+import androidx.core.net.toUri
+
+/**
+ * Handles exporting and importing the internal Room database, DataStore settings,
+ * and encrypted tokens. Bundles everything into a single ZIP archive.
+ *
+ * Exporting uses a checkpoint-then-zip approach to ensure WAL-mode
+ * consistency. Importing replaces the database and preferences files on disk
+ * and requires an app restart to take effect.
+ */
+@Singleton
+class DatabaseBackupManager @Inject constructor(
+    @ApplicationContext private val context: Context,
+    private val database: StashDatabase,
+) {
+
+    /**
+     * Exports the database and settings to the provided [targetUri] as a ZIP.
+     */
+    suspend fun exportDatabase(targetUri: Uri): Result<Unit> = withContext(Dispatchers.IO) {
+        runCatching {
+            // 1. Force a checkpoint to ensure the .db file is up to date
+            database.openHelper.writableDatabase.query(
+                SimpleSQLiteQuery("PRAGMA wal_checkpoint(FULL)")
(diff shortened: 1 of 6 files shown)


rawnaldclark:
Hey @winstriker — thanks for this, really thorough work and the WAL-checkpoint-then-zip approach is the right pattern. The SAF re-grant flow was a nice catch too.

Two things I'd want to lock down before merging, plus a note on the conflicts:

**1. Zip Slip guard.** In `DatabaseBackupManager.importDatabase` the entry name is used directly:

```kotlin
val file = File(datastoreDir, entry.name.substringAfter("datastore/"))
```

A crafted entry like `datastore/../../somewhere/evil` would escape `datastoreDir`. Could you add a canonical-path check before writing each entry? Roughly:

```kotlin
if (!outFile.canonicalPath.startsWith(datastoreDir.canonicalPath + File.separator) &&
    outFile.canonicalPath != dbFile.canonicalPath
) {
    throw SecurityException("Entry escapes target directory: ${entry.name}")
}
```

**2. Validate the ZIP is actually a Stash backup.** Right now if a user picks any random ZIP, the import iterates entries and writes whatever's inside them, potentially destroying their library. Could you add a small manifest entry to the zip — say `manifest.json` with `{ "versionCode": N, "dbSchemaVersion": M, "exportTimestamp": T }` — and refuse import if it's missing or if `dbSchemaVersion` is newer than the current schema? That gives us both a "is this even a Stash backup" check and a "is this from a future version" safety net.

**3. Conflicts.** The Settings files conflict with v0.9.30 (just shipped — Online streaming engine added a Playback section + injected `StreamingPreference` into `SettingsViewModel`). Both your changes and mine are additive in different parts of the file, so it should rebase cleanly. Could you `git rebase origin/master` and push?

One small thing for a follow-up (not a blocker): the regex extracting the SAF tree URI from the raw DataStore Preferences file is fragile — DataStore is protobuf-serialized, so it only works because SAF URIs happen to be ASCII and survive a UTF-8 decode of protobuf bytes intact. Would be more robust to actually deserialize the preferences file (you'd need to read it with the same DataStore instance after restore and consult the SAF key directly), but it works today.

Thanks again — happy to merge once those two land.

winstriker:
Hi @rawnaldclark, I've implemented the small fixes you proposed and removed the SAF URI regex extraction. 

As i was testing, offline playback after updating to v9.30.0 was dead when clicking on a single song that was downloaded in an external folder.
I pushed a fix in f40173a that also checks for files in SD/OTG locations, i think it would close #75.

Let me know if there is something else to improve!

rawnaldclark:
Audit complete — everything checks out:

- Zip Slip guard handles all escape vectors I could throw at it (`../`, absolute paths, mixed separators).
- Manifest validation correctly runs **before** `database.close()`, so a bad ZIP can't disturb live state.
- The tmp-file dance for the post-restore DataStore peek is clever — sidesteps the "only one DataStore per file" singleton constraint properly.
- The offline regression fix in `filePathExistsOnDisk` is the right root-cause fix. Applied consistently in all three call sites (`PlayerRepositoryImpl`, `deleteTrackFile`, orphan cleanup).
- Build passes on the rebased branch; my v0.9.30 Playback section survived the rebase intact.

Merging now. Thanks again — this is solid work, and the offline fix is a real save. Shipping as v0.9.31.

PR #148, opened 2026-05-31
feat: Last.fm Worker proxy + shared cache (durable rate-limit fix)

rawnaldclark:
The durable fix for the shared-key Last.fm rate limit (error 29) that breaks custom Stash Mixes. Stacks on #146 — **review/merge #146 first** (this PR's app-side change builds on its `unsignedGet`).

### The Worker (`infra/lastfm-proxy/`)
Proxies the **generic, cross-user** Last.fm read lookups (`tag.getTopTracks`, `artist.getSimilar`, etc.) through **one server-side key** with an **edge cache**. Each unique query hits Last.fm at most once per TTL and is served to every user from cache — so Last.fm request volume is **decoupled from user count**, and the key can't get throttled regardless of growth.

- `GET /lastfm?method=…` — method-allowlisted generic reads only; rejects `api_key`/`api_sig`/`sk`/`user`/`username` (never signed, never per-user).
- Injects its own key (Worker secret `LASTFM_API_KEY`), caches via the edge Cache API (14-day TTL), returns 429 on upstream error 29.
- Mirrors `infra/tipjar-worker`. `node:test` logic tests (cache-key normalization, error parsing, allowlist) — 5 green. README has deploy + wiring steps.

### App toggle
`LastFmCredentials.proxyUrl` (← `lastfm.proxyUrl` / `BuildConfig.LASTFM_PROXY_URL`). When set, `unsignedGet` routes cacheable reads through the Worker (strips api_key; breaker keyed `__proxy__`); auth/scrobble/per-user reads stay direct. **Blank by default = no-op** until you deploy + set the URL.

### Personalization is unchanged
The proxy serves only the generic candidate *pool*; seed selection, scoring and filtering stay on-device per user. The on-device cache + per-key breaker (from #146) act as L1/fallback in front of the Worker.

### To deploy
`infra/lastfm-proxy/README.md` — `wrangler login`, `wrangler secret put LASTFM_API_KEY` (use a dedicated key), `wrangler deploy`, then set `lastfm.proxyUrl`. (Deployment needs your Cloudflare account — can't be done from CI here.)

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -70,6 +70,11 @@ val lastFmApiSecret: String =
 // primary key above. Empty by default.
 val lastFmExtraApiKeys: String =
     localProperties.getProperty("lastfm.extraApiKeys") ?: System.getenv("LASTFM_EXTRA_API_KEYS").orEmpty()
+// Optional Stash Last.fm proxy Worker base URL. When set, generic read lookups
+// route through it (the Worker holds the server key + shared cache), decoupling
+// Last.fm load from user count. Empty = app talks to Last.fm directly.
+val lastFmProxyUrl: String =
+    localProperties.getProperty("lastfm.proxyUrl") ?: System.getenv("LASTFM_PROXY_URL").orEmpty()
 
 android {
     namespace = "com.stash.app"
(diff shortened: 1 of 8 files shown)


rawnaldclark:
Merged into master via the 0.9.41 release (commits landed directly). Superseded by tag v0.9.41.

Issue #154, opened 2026-06-02
Android Auto - playlist


CodeF00l:
When choosing a song from a playlist on Android Auto in the car, it only plays the song. It doesn't queue the playlist. 

A workaround is to press on shuffle.

rawnaldclark:
The fix for this is now merged: tracks browsed inside an Android Auto playlist carry their playlist context again, so tapping a song queues the WHOLE playlist starting at that song (next/previous work, and an in-order tap plays in order — shuffle stays on the Shuffle Play entry). It will be in the next release. Since I verified the queue logic on-device but not on a car head unit, a confirmation from your car once you update would be very welcome.

CodeF00l:
Confirmed this works today, thanks for the support!

PR #160, opened 2026-06-04
Fix Bluetooth/Android Auto resume (full queue + position), add Resume shortcut & test CI

### LINKED ISSUE #155: Android auto - autoplay
When turning on the car and loading android auto, the app will automatically play the latest downloaded song and not the last played.

This happens when the app newly boots. I.e. If you prepare a playlist or a song on the phone it'll autoplay that.

More of a minor annoyance :)

midlan:
## TL;DR

Fixes Bluetooth / Android Auto **playback resumption**, which previously replayed the *wrong* track as a one-song queue.

- **Fixes #155** — on a cold start, resume now continues the **last-played queue at the saved track and position** instead of playing the most-recently-downloaded song.
- **next / previous work after resuming** — the whole queue is restored, not a single item.
- **Playback actually starts** — both from a warm process (it used to load but stay paused) and from a cold start.
- **Works in online (streaming) mode too**, not just downloaded files.
- Adds a **"Resume playback" app shortcut** (launcher long-press + Samsung Bixby Routines) that reliably resumes from a **cold start and the lock screen** — the generic "play music" media-button action can't.
- Adds **GitHub Actions** (unit tests + a debug-APK artifact) so contributors without a local Android build/test environment can still get CI feedback and an installable build.

Single squashed commit, rebased on current `master`. `core:media` unit tests and the debug APK build green in CI; the resume + shortcut behaviour was verified on-device (Samsung / Android Auto / car Bluetooth).

---

## The bug (#155)

`StashPlaybackService.onPlaybackResumption()` returned a **single** `MediaItem` with `startPositionMs = C.TIME_UNSET`:

- It fetched only the last-played track (falling back to the most-recently-*added* track), so a cold boot played the newest download — exactly what #155 describes.
- The returned queue had **one** item, so next/previous had nowhere to go.
- The saved position was discarded, so it always restarted from 0:00.

`PlaybackStateStore` already persisted `trackId` / `positionMs` / `queueIndex`, but nothing read them back, and the **queue list itself was never persisted** — so even the saved index was meaningless.

### 1. Restore the full queue, track and position
- `PlaybackStateStore` now also persists the ordered **queue track ids** + shuffle state (`saveQueue`), written from `PlayerRepositoryImpl.updateState()` only when the queue/shuffle actually changes (not on every position tick).
- `PlaybackResumer` (new, unit-tested) rebuilds the ordered queue, locates the saved current track by id (robust to deleted rows), and returns the start index + position + shuffle.
- `onPlaybackResumption` hands Media3 the **whole queue** at the saved index/position; the old single-track behaviour remains only as a fallback when nothing has been persisted yet.

### 2. Make playback actually start
From a warm process the queue would load but stay **paused** until a second trigger. Media3 is supposed to auto-play after `onPlaybackResumption` on a real play request, but that didn't reliably take. `ResumePlayGate` (new, unit-tested) arms on a genuine play request (`isForPlayback`) and forces `play()` once the restored queue lands — while staying inert for boot-time notification population, so the device never auto-plays after a reboot.

### 3. Online (streaming) mode
The service's player uses the default media-source factory, so a streamed track needs a resolved `http` URL before it reaches the player (normally done in `PlayerRepositoryImpl.setQueue`, which isn't in the loop on the native resume path). `ResumeStreamResolver` (new, unit-tested) resolves the current track's stream URL on resume, reusing the existing `StreamSourceRegistry` / `StreamUrlCache` and the same streaming gates (pref / connectivity / cellular). Because that resolve can be slow and this path is started via `startForegroundService`, a brief **"Resuming…"** foreground notification is posted to satisfy the OS 5-second `startForeground` deadline, reusing Media3's notification id so the real media notification replaces it seamlessly.

### 4. "Resume playback" app shortcut (the cold-start / lock-screen path)
A routine that fires the generic **"play music"** media-button at a *dead* app can't reliably wake it — Android won't deliver media buttons to a not-running (or force-stopped) process. VLC works around this with a dedicated **app shortcut**: launching an explicit activity cold-starts the app and works behind the lock screen, unlike a media key.

This PR adds the same: a static `shortcuts.xml` → `ResumePlaybackActivity` (a no-UI trampoline that calls `PlayerRepository.resumeLastQueue()` and finishes). `resumeLastQueue()` reuses the normal `setQueue` path, so it resolves + plays in **both** offline and online modes with the existing background queue-fill. The shortcut appears on launcher long-press and is selectable as a **Bixby Routine** action — point the car routine at it instead of "play music".

> Note on force-stop: if the app is **force-stopped** from Settings, Android puts it in the stopped state and won't deliver media buttons or start it for resumption until it's manually launched again — that's an OS guarantee no app can override. Normal process death (memory reclaim, swipe-away) is unaffected.

### 5. GitHub Actions for contributors without a local build/test env
The repo previously had only `release.yml` (tag-triggered APK build). This adds `tests.yml` (on push / PR):
- **`core:media` unit tests** — fast feedback on the playback logic.
- **A debug APK artifact** — so a contributor (or tester) without Android Studio / a build environment can download an installable build straight from the run.

> The debug APK carries the ~145 MB yt-dlp/Python bundle, so the artifact is large; feel free to trim that job or shorten retention if it's too heavy for the project's Actions usage.

## Testing
- New unit tests: `PlaybackResumerTest`, `ResumePlayGateTest`, `ResumeStreamResolverTest`; existing `PlayerRepositoryStreamingTest` updated for the new constructor. All `core:media` tests pass in CI.
- Manually verified on a Samsung device with Android Auto / car Bluetooth: correct track + position on resume, working next/previous, playback starting from both warm and cold states, and the shortcut firing from a cold/locked phone.
- The trampoline-activity → `resumeLastQueue` → `setQueue` wiring needs a live `MediaController`, so it's covered by the on-device test rather than a unit test.

## Scope notes
- Only the **current** track is foreground-resolved on the native car/AVRCP resume path; subsequent streamed tracks resolve via the existing prefetch/skip path. The shortcut path background-fills the whole queue, so it's the stronger option for online next/previous from cold.
- Does **not** address #154 (tapping a song in the Android Auto browser not queueing the playlist) — that's a separate browse-tap code path.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,83 @@
+name: Unit tests
+
+# Runs JVM unit tests on push and PR. The repo previously had no test CI —
+# only release.yml, which builds the APK on `v*` tags. This validates the
+# media module (playback resume, queue, streaming) on every change and also
+# builds a debug APK artifact for manual on-device testing.
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  unit-tests:
+    name: core:media unit tests
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: '17'
+
+      - name: Set up Gradle cache
+        uses: gradle/actions/setup-gradle@v4
+
+      - name: Fix gradlew permission
+        run: chmod +x gradlew
+
+      - name: Run core:media unit tests
+        run: ./gradlew :core:media:testDebugUnitTest --no-daemon
+
+      - name: Upload test report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: core-media-test-results
+          path: |
+            core/media/build/test-results/testDebugUnitTest/
+            core/media/build/reports/tests/testDebugUnitTest/
+
+  build-debug-apk:
+    name: Build debug APK
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: '17'
+
+      - name: Set up Gradle cache
+        uses: gradle/actions/setup-gradle@v4
+
+      - name: Fix gradlew permission
+        run: chmod +x gradlew
+
+      - name: Assemble debug APK
+        run: ./gradlew :app:assembleDebug --no-daemon
+
+      - name: Locate debug APK
+        id: find_apk
+        run: |
+          APK=$(find app/build/outputs/apk/debug -name "*.apk" | head -n1)
+          if [ -z "$APK" ]; then
+            echo "::error::No debug APK produced by assembleDebug"
+            exit 1
+          fi
+          echo "apk_path=$APK" >> "$GITHUB_OUTPUT"
+          echo "Built: $APK ($(du -h "$APK" | cut -f1))"
+
+      - name: Upload debug APK
+        uses: actions/upload-artifact@v4
+        with:
+          name: stash-debug-apk
+          path: ${{ steps.find_apk.outputs.apk_path }}
--- a/app/src/debug/res/xml/shortcuts.xml
+++ b/app/src/debug/res/xml/shortcuts.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shortcuts xmlns:android="http://schemas.android.com/apk/res/android">
+    <!--
+        Debug-variant override of app/src/main/res/xml/shortcuts.xml. Identical
+        except targetPackage carries the ".debug" applicationIdSuffix so the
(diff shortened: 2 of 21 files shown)


midlan:
Just force pushed icon fix, it was white on white launcher background. Now stash purple is used around the play icon of resume shurtcut.

image

rawnaldclark:
﻿Merged — thank you for an outstanding PR. The queue/position persistence design, the ResumePlayGate boot-safety distinction, and the test extraction made this easy to trust, and your on-device verification held up: I re-verified cold-start resume (process killed via stop-app, resumed via the shortcut) with full queue + working next/previous on a Pixel today.

Two adaptations during integration, since master had moved under you (v0.9.50 reworked the queue display):
1. The queue persistence in `updateState()` now saves the **logical** queue (`newState.queue`) — the raw-timeline local your branch read no longer exists, and the timeline is a sparse resolved-only subset during streaming, so persisting it would have resumed a partial queue.
2. `tests.yml` push trigger is scoped to master so PR branches do not double-run and upload a ~145 MB debug APK per push — PR runs still cover contributors.

Also landed alongside it: the Android Auto browse-tap fix (#154/#173) — browse children now carry their parent playlist id so a tap queues the whole playlist, reusing your test-extraction pattern (`AutoBrowseQueue`). Both ship in the next release.

PR #172, opened 2026-06-07
 fix(lastfm): detect repeat-one loop

ParaliyzedEvo:
## Summary
Repeat-one scrobbles were silently dropped because the loop detection collected playerState instead of currentPosition, missing position resets entirely. 

## Changes
Added RepeatMode import
Added Collector 2 for repeat-one loop detection using playerRepository.currentPosition on Dispatchers.Main
Extracted schedulePendingFire() shared helper so both collectors reuse the same scheduling logic
lastPositionMs moved from class-level to local variable inside Collector 2's coroutine
Added KDoc invariant documenting the repeat-one behavior

Quick note: Again, debug.yml is not needed for merge as it was for my testing.


rawnaldclark (changes requested):
﻿Thanks for this — repeat-one plays absolutely should scrobble each loop, and routing both paths through a shared `schedulePendingFire` is the right shape. A few things need fixing before merge:

**1. `lastPositionMs` never resets on track change (real bug).** It only resets when `currentTrack` is null. Sequence: listen to track A past 10s with repeat-one on, manually switch to track B → first position tick for B is near 0 while `lastPositionMs` is still large → `positionJumpedBack` fires and `schedulePendingFire` runs for B, racing the track-change collector that just scheduled B (double now-playing, reset threshold timer). Track the track id alongside the position and reset `lastPositionMs` whenever the id changes.

**2. Seek-back counts as a new play (doc/code mismatch).** The KDoc says "a position reset back to near zero", but the code only checks `positionMs < lastPositionMs - 5_000`. A user scrubbing from 3:00 back to 2:00 with repeat-one on gets a fresh scrobble session — that inflates scrobbles. Add the near-zero check the doc already promises (e.g. `positionMs < 5_000`), which distinguishes an actual loop restart from a scrub.

**3. `Log.d` on every position tick.** The `repeat-check` log fires once per position update for every user, forever. Please drop it (or log only inside the `if`).

**4. `debug.yml` overlaps with #160's `tests.yml`**, which also builds a debug APK artifact (and runs unit tests). Building on every push to every branch also burns Actions minutes — the debug APK with the bundled yt-dlp/Python runtime is ~145 MB per artifact. Prefer dropping the workflow from this PR and letting the CI land via #160.

**5. Tests.** The repo has `ListeningRecorder` test coverage nearby (`ListeningRecorderSkipTest`); a unit test for the loop-detector (loop restart fires, seek-back does not, track-change resets state) would lock this in.

Also nit: both touched files lost their trailing newline, and several explanatory comments in `start()` were removed — please restore those.

ParaliyzedEvo:
@rawnaldclark Made all the changes necessary, testing works and everything looks good to me as it passed both checks and complied

ParaliyzedEvo:
@rawnaldclark added a timeout on the test.yml file as it seemed to hang quite a bit on the unit tests

rawnaldclark:
Taking a look tomorrow 👍👍

ParaliyzedEvo:
@rawnaldclark kk thanks, while im at it, imma try to look into why the check is just hanging.
Edit: Sooooooo I actually dont know what the problem is (again new to Koltin), like I've tried increasing the max heap size to 2g and it still hanged in my repo, but maybe you have a fix?

rawnaldclark:
Took this over the finish line with one maintainer fixup commit (thanks for the fix + the two rounds!):

- **Kept** the actual fix — `ListeningRecorder` repeat-one loop detection (shared `schedulePendingFire()`, near-zero guard, track-change reset) and the `currentPosition.flowOn(Dispatchers.Main)` in `PlayerRepositoryImpl` (required — the controller is main-thread-only).
- **Restored** the two skip tests the rename dropped (skip-before-threshold / listen-no-skip-after) alongside your three repeat tests — skip coverage feeds MixGenerator's skip penalty. Both now share one position-aware fake.
- **Reverted** the CI-hang stabs (the `tests.yml` `--debug … || true` + LASTFM secrets, `maxHeapSize=2g`, `.gitignore build-logic/`). The failing `:core:media:testDebugUnitTest` job is a known flaky network-test hang that hits master too — not your bug and not heap-fixable. The **Build debug APK** job is the real signal.
- Fixed the missing trailing newline.

All five tests in `ListeningRecorderSkipTest` pass locally. Merging.

PR #215, opened 2026-06-25
UI polish: mini-player visibility, AMOLED theme, scrollbars + streaming fixes

ParaliyzedEvo:
## Summary
- **Mini-player visibility logic** — mini-player now hides on Settings/Account
  sub-pages and other utility-only inner pages (Equalizer, Library Health,
  Blocked Songs, Failed Matches, Failed Downloads), while keeping the bottom
  nav bar visible. Also hides during active web-login flows and on the
  squid.wtf captcha screen. Content pages (playlists, artists, albums, liked
  songs) keep the mini-player visible since music is actively playing there.
- **AMOLED theme** — new pure-black theme option with a matching ambient
  background mode on Now Playing (optional blur-layer toggle).
- **Draggable scrollbars** — fade-in/fade-out scrollbar with a narrow
  drag-handle hit zone added to Library's grids/lists and Playlist Detail,
  without blocking normal scroll/tap gestures.
- **Streaming + download feedback** (credit: @TENDIEMANCOMETH (discord)/
  @JoshDui) — improved user-facing feedback during slow stream resolves
  and fixed download-state reset issues.
- **Crossfade duration slider** — fixed a rounding bug where dragging the
  slider skipped/repeated values instead of landing on clean 1s increments.
- **Streaming source reliability** — stream resolution now skips lossless
  sources whose build-time credentials aren't configured instead of wasting
  a network round-trip on them, and no longer double-hits parked (host-down)
  proxies during normal playback. Also fixes a large-library sync
  performance issue where syncing thousands of new tracks could take
  several seconds and spam logcat.
  
### CI / tooling
- `.github/workflows/release.yml`, `.github/workflows/tests.yml` — bumped
  Actions to Node 24 (`actions/checkout@v6` etc.)
- `.jabbarc` — pin Jabba's JDK version for CI consistency
- `.vscode/settings.json` — editor settings for contributors using VS Code
- `.github/workflows/auto-label.yml` — added auto-labeling for the bug report template: labels issues by the "Where does it happen?" area, auto-creates the area labels if they don't exist yet, and adds a manual `workflow_dispatch` trigger to backfill labels onto existing untagged issues
- `.github/workflows/tests.yml` — skip the test job on push when the
  commit message contains `[skip test]`
- `.github/dependabot.yml` (new) — daily dependency update checks,
  ignoring major-version bumps

### Mini-player visibility
- `app/.../navigation/StashScaffold.kt` — reworked bottom-chrome logic so
[...]

PR #299, opened 2026-07-18
Improve streaming reliability, sync performance, and CI tooling

ParaliyzedEvo:
## Summary
- **CI / tooling** — modernized the project's automation and developer
  tooling by updating GitHub Actions, adding Dependabot and automatic issue
  labeling, improving CI consistency, and adding contributor-focused editor
  settings.
- **Crossfade duration slider** — fixed a rounding bug where dragging the
  slider skipped/repeated values instead of landing on clean 1s increments.
- **Streaming source reliability** — stream resolution now skips lossless
  sources whose build-time credentials aren't configured instead of wasting
  a network round-trip on them, and no longer double-hits parked (host-down)
  proxies during normal playback. Also fixes a large-library sync
  performance issue where syncing thousands of new tracks could take
  several seconds and spam logcat.
  
### CI / tooling
- `.github/workflows/release.yml`, `.github/workflows/tests.yml` — bumped
  Actions to Node 24 (`actions/checkout@v6` etc.)
- `.jabbarc` — pin Jabba's JDK version for CI consistency
- `.vscode/settings.json` — editor settings for contributors using VS Code
- `.github/workflows/auto-label.yml` — added auto-labeling for the bug report template: labels issues by the "Where does it happen?" area, auto-creates the area labels if they don't exist yet, and adds a manual `workflow_dispatch` trigger to backfill labels onto existing untagged issues
- `.github/workflows/tests.yml` — skip the test job on push when the
  commit message contains `[skip test]`
- `.github/dependabot.yml` (new) — daily dependency update checks,
  ignoring major-version bumps

### Crossfade duration slider
- `feature/settings/.../SettingsPlaybackScreen.kt` — round instead of
  truncate the slider's float value so dragging lands on every integer
  step (1–12s) instead of skipping some and repeating others
  
### Streaming source reliability
- `data/download/build.gradle.kts` — expose per-source `QBDLX_CONFIGURED`
  and `ARCOD_CONFIGURED` BuildConfig flags so an unconfigured source is
  skipped independently rather than gating the whole lossless chain
- `data/download/.../lossless/LosslessSourceRegistry.kt` — skip
  unconfigured sources up front on the download-resolve path
- `core/media/.../streaming/StreamSourceRegistry.kt` — same gating for
  streaming resolves; also removed a leftover pair of live (uncommented)
  kennyy/squid calls that were still hitting those parked, host-down
  proxies on every stream resolve despite being documented as disabled,
  and removed duplicate qbdlx/amz entries in the resolver chain

### Sync performance
- `core/data/.../sync/workers/DiffWorker.kt` — batches new-track discovery
  during sync into a handful of bulk DB operations instead of one
  round-trip per track; large syncs (thousands of new tracks) previously
  triggered multi-second single transactions and could produce
  `CursorWindow`/out-of-memory warnings under log pressure
- `core/data/.../db/dao/TrackDao.kt`,
  `core/data/.../db/dao/PlaylistDao.kt` — added batched lookup/insert
  DAO methods backing the above
 
## How was this tested?
- [x] `./gradlew test` passes

### DIFF
--- a/core/media/src/main/kotlin/com/stash/core/media/streaming/StreamSourceRegistry.kt
+++ b/core/media/src/main/kotlin/com/stash/core/media/streaming/StreamSourceRegistry.kt
@@ -3,6 +3,7 @@ package com.stash.core.media.streaming
 import android.util.Log
 import com.stash.core.data.db.entity.TrackEntity
 import com.stash.core.data.prefs.StreamingPreference
+import com.stash.data.download.BuildConfig  
 import javax.inject.Inject
 import javax.inject.Singleton
 
@@ -81,11 +82,16 @@ class StreamSourceRegistry @Inject constructor(
      *   via the fast InnerTube engine only (no slow yt-dlp). Used by the
      *   background-fill path so a 15-35s yt-dlp invocation never sits on
      *   the queue's critical path. Foreground calls leave this true.
+     * @param preferFastStartup when true, skip slow job/full-file lossless
+     *   fallbacks on the normal branch and reach YouTube sooner. Used for
+     *   cellular foreground/next-up resolves. Force toggles still override
+     *   this policy so diagnostics can exercise one source.
      */
     suspend fun resolve(
         track: TrackEntity,
         allowYouTube: Boolean = true,
         allowYtDlp: Boolean = true,
+        preferFastStartup: Boolean = false,
     ): StreamUrl? {
         val resolvers = buildList<Pair<String, suspend (TrackEntity) -> StreamUrl?>> {
             if (streamingPreference.isForceQbdlxOnly()) {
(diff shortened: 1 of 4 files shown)


rawnaldclark (commented):
Thanks @ParaliyzedEvo — audited the whole PR. Most of it we want immediately; one part needs to come out, and one needs tests. Proposal: **split it**.

**Merge-ready today (happy to take as its own PR):** the Actions version bumps (release.yml is surgical and keeps the token-fetch step intact 👍), Dependabot, issue auto-labeling, `.jabbarc`/`.vscode`, the crossfade slider `roundToInt()` fix, and the LosslessSourceRegistry unconfigured-credentials skip (nice touch for fork builds).

**Needs tests before merge:** the DiffWorker batching. The concept is right — the bulk cross-ref insert and the batched identity lookup kill thousands of per-track SELECTs — but this rewrites ~230 lines of the most correctness-sensitive path in the app (REFRESH vs ACCUMULATE semantics, the soft-delete guard, `addedAt` preservation). That code has bitten us before. Please add coverage for: a REFRESH diff that must not resurrect a soft-deleted track, an ACCUMULATE diff preserving `addedAt` on re-stamped rows, and the batch identity match honoring the spotifyUri > youtubeId > canonical priority when multiple candidates hit. Also note the sentinel contract on `findExistingForBatch` deserves a test since it's a footgun.

**Please drop:** `preferFastStartup`. It's currently dead (no caller passes `true`), and its intended behavior — skipping lossless sources on cellular — is the exact regression that kept #215's streaming half out: lossless never tried on cellular + lossy results poisoning the shared StreamUrlCache (which keys by trackId with no quality/network dimension, so a cellular lossy URL gets served later on WiFi). If fast cellular startup is worth pursuing, let's design it together first — it needs a cache guard and a background lossless upgrade path to be shippable.

Split it that way and the first PR can merge same-day.

rawnaldclark:
Post-merge state, for the record: both reverts (357be35b DiffWorker batching, 9728c90f fast-startup + config gates) verified clean against pre-PR code — master compiles, suites green. What remains from this PR on master: the CI/Actions bumps, Dependabot, auto-labeling, editor configs, and the crossfade slider fix — all good. Standing invitations from the review: (1) the QBDLX/ARCOD_CONFIGURED source gates were a genuinely good idea for fork builds and got swept out with the fast-startup revert — happy to merge them same-day as their own small PR; (2) the DiffWorker batching perf win is real and welcome back with the soft-delete/addedAt/identity-priority tests described above. One process ask: let's route changes through PRs with a review rather than direct pushes to master — it keeps the squash history clean and saves us colliding mid-fix (we crossed wires twice tonight).

PR #302, opened 2026-07-19
chore(streaming): remove the dormant preferFastStartup parameter

rawnaldclark:
Follow-up to the #299 review: the parameter was merged caller-less (no
call site passes true, behavior unchanged) and its intended policy —
skipping lossless sources on cellular — is the exact pattern that kept
PR #215's streaming half out: lossless never tried on cellular, and the
lossy results poisoned the shared StreamUrlCache (keyed by trackId with
no quality/network dimension). Dead code + a loaded gun; removed.

A comment on the amz resolver records the constraint for whoever picks
up fast-startup properly: it needs a cache quality/network guard and a
background lossless-upgrade path first.

Keeps everything else #299 landed: the unconfigured-source gates, the
arcod re-enable behind ARCOD_CONFIGURED, CI/tooling, and the crossfade
slider fix.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


rawnaldclark:
Superseded — ParaliyzedEvo reverted the fast-startup commit on master directly (9728c90f), which covers this PR's purpose. Note the revert also removed the QBDLX/ARCOD_CONFIGURED gates and re-parked arcod; those are welcome back as a clean split PR per the #299 review.

PR #327, opened 2026-07-20
fix: add artwork to liked playlists

### LINKED ISSUE #67: Liked songs background art missing in library playlist
Liked songs spotify and liked in app... background art missing in library playlist

arpitagarwal1301:
## Summary

- seed Spotify and YouTube liked-playlist covers from the first available track artwork
- backfill missing external liked-playlist artwork without replacing established covers
- seed or repair the in-app Liked Songs cover from the liked track while preserving later artwork

## Validation

- focused `:core:data:testDebugUnitTest` coverage for artwork selection, sync backfill, and local-liked seeding: 12 tests passed
- `:feature:library:testDebugUnitTest`: 57 tests passed
- `assembleDebug`: passed
- `git diff --check`: passed
- five unrelated core-data failures reproduce identically on clean `origin/master`

## Device testing

- Not run. This is a data-path fix and the existing Library playlist-card UI was unchanged; authenticated Spotify/YouTube seed data was unavailable.

Fixes #67

### DIFF
--- a/core/data/src/main/kotlin/com/stash/core/data/social/stash/StashLikedPlaylistRepository.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/social/stash/StashLikedPlaylistRepository.kt
@@ -38,7 +38,11 @@ class StashLikedPlaylistRepository @Inject constructor(
      * leaves the existing timestamp untouched.
      */
     suspend fun add(trackId: Long) {
-        val playlistId = ensureSeeded()
+        val trackArt = trackDao.getById(trackId)?.let { track ->
+            sequenceOf(track.albumArtPath, track.albumArtUrl)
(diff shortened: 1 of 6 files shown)


rawnaldclark:
Merged with the rest of your batch — thanks. One finding from on-device testing worth a follow-up: on an account whose liked songs haven't changed, DiffWorker's unchanged-snapshot fast path skips processPlaylist entirely, so the LIKED_SONGS art backfill in finalizePlaylistMetadata doesn't run until the next real change to the liked set. Verified on a device where all three liked playlists stayed NULL through a full sync (last_synced never updated for them). Not a defect in your change — the backfill is correct once reached — but if you want a v2, hoisting the missing-art repair ahead of the short-circuit would make it fire on the first sync after update. Happy to take that as a small follow-up PR.

PR #345, opened 2026-07-21
Parse the flat InnerTube search shape so results appear for all regions

### LINKED ISSUE #268: [Bug]: Search for artists doesn't work
### Before you start

- [x] I [searched existing issues](https://github.com/rawnaldclark/Stash/issues?q=is%3Aissue) and this isn't a duplicate

### Stash version

v.0.9.75

### Phone & Android version

Huawei nova 11, android 12

### Where does it happen?

Search

First, the search for an artist had never worked without VPN for me. Now it doesn't even with VPN turned on. (But Spotify playlists still work).

### Steps to reproduce

1. Open the search tab
2. Search for any artist
3. Can't find any artist. (Previous saved searches remain in the panel. But it opens an empty page).

Image

Image

Image

chaudhary-lakshay:
Closes #268

### Root cause

The reporter (and a commenter) suspected regional blocking, but the app already pins `gl=US` / `hl=en` on every InnerTube request — so I captured a live response with the exact context the app sends. YouTube serves **two search response shapes**:

- **Titled shape** (authenticated YT-Music clients in supported regions): `sectionListRenderer.contents` holds `musicShelfRenderer` shelves titled "Songs" / "Artists" / "Albums". This is the only shape `searchAll` parsed.
- **Flat shape** (unauthenticated or region-limited requests — the reporter's case): the same path instead holds a run of `itemSectionRenderer` wrappers, each carrying one `musicResponsiveListItemRenderer` row, with **no titled shelves at all**.

On the flat shape `searchAll` found no `musicShelfRenderer`, produced zero sections, and the UI rendered "No results found" — over a response full of hits. My captured `bjork` response in that shape contained 11 songs, 3 artists, and 3 albums, all silently discarded. This also explains why VPN location changes toggled the bug: the account/IP combination decides which shape you get, not whether results exist.

### Fix

- **`FlatSearchResponseParser.kt`** (new): walks the flat `itemSectionRenderer` rows and classifies each by its navigation target — a `videoId` → song, `MUSIC_PAGE_TYPE_ARTIST` → artist, `MUSIC_PAGE_TYPE_ALBUM` → album — then groups them into the existing `SearchResultSection` types (songs capped at 4, matching the titled-shelf cap).
- **`searchAll`**: falls back to the flat parser **only when no titled list shelf matched**, so the authenticated path is untouched.
- **`parseTrackSummaryFromListItem`**: flat song rows subtitle as `"Song • <artist>"`; a single leading content-type label is now dropped so it can't pollute the artist string. Legacy titled-shelf rows carry no such label and are unaffected (existing separator filtering already covered `" • "`).

The top-result card (`musicCardShelfRenderer`) is present in both shapes and keeps working unchanged.

### Testing

- Captured a real flat-shape response from the live endpoint and validated the classification extracts all 17 entities before writing the Kotlin.
- `FlatSearchResponseParserTest` (new): 5 tests over a fixture mirroring the real shape — section grouping, song label-stripping, artist browseId, album artist/year subtitle parsing, empty input.
- `:data:ytmusic:compileDebugKotlin` clean; new tests pass.
- Pre-existing note: the whole `YTMusicApiClientTest` class currently fails on `master` in my environment with `InvalidUseOfMatchersException` (a Mockito matcher misuse in the test's own fake) — unrelated to this change, identical failure count with the change stashed.

### Possible follow-up

The issue's second screenshot (tapping a saved artist → empty profile page) goes through the artist `browse` path, which may need the same flat-shape treatment — I kept this PR scoped to search. Happy to look at that separately if wanted.

### DIFF
--- a/data/ytmusic/src/main/kotlin/com/stash/data/ytmusic/FlatSearchResponseParser.kt
+++ b/data/ytmusic/src/main/kotlin/com/stash/data/ytmusic/FlatSearchResponseParser.kt
@@ -1,0 +1,152 @@
+package com.stash.data.ytmusic
+
+import com.stash.data.ytmusic.model.AlbumSummary
+import com.stash.data.ytmusic.model.ArtistSummary
+import com.stash.data.ytmusic.model.SearchResultSection
+import com.stash.data.ytmusic.model.TrackSummary
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonObject
+
+/**
+ * Parser for the "flat" InnerTube search response shape.
+ *
+ * The titled `musicShelfRenderer` shape ("Songs"/"Artists"/"Albums") that
+ * [parseSongsShelf] / [parseArtistsShelf] / [parseAlbumsShelf] expect is only
+ * served to authenticated YouTube-Music clients in supported regions. In other
+ * cases — unauthenticated, or a region where the request is treated as plain
+ * YouTube — InnerTube returns a FLAT list instead: `sectionListRenderer.contents`
+ * holds a run of `itemSectionRenderer` wrappers, each carrying a single
+ * `musicResponsiveListItemRenderer` row, with no titled shelves at all.
+ *
+ * The legacy parser finds no `musicShelfRenderer` in that shape and yields zero
+ * sections, so the Search screen shows "No results found" even though the
+ * response is full of real songs/artists/albums (issue #268). This parser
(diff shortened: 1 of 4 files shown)


rawnaldclark:
Merged — this was a latent bug we'd had flagged internally since the discoverability work, and your live-capture diagnosis nailed the mechanism. The only-when-titled-parse-is-empty gate is exactly the right regression posture. Ships in v0.9.81 within the hour.

PR #346, opened 2026-07-21
Never classify files on an unmounted volume as MISSING in the repair sweep

### LINKED ISSUE #98: Last Storage Location
The issue is simple I use a Micro SD card for storing the music tracks, the songs where properly synced up showing in the app in the library section, but I when I removed my Micro SD card and again installed it but now the whole Library section is empty.

chaudhary-lakshay:
Fixes #98

### Scope note

The report predates the current download-integrity sweep (the reporter was on ~v0.9.36, May 24; the sweep landed June 3 in `5afb6ab8`), so I can't prove which code path emptied their library at the time. What I can prove is that **current `master` has exactly this data-loss trap**, and this PR closes it — so this PR is best read as "prevents this class of loss" - if the maintainer prefers to keep #98 open for the historical report, feel free to unlink.

### The trap

`reconcileMissingDownloadedFiles()` runs on **every launch** and un-marks (`bulkResetForReDownload`: `is_downloaded = 0`, `file_path` nulled) every row whose file classifies as `MISSING`. In `LocalFileOps.classify()`, the plain-path branch was:

```kotlin
!f.exists() -> LocalFileState.MISSING       // internal storage: reliable
```

The comment's assumption doesn't hold: plain paths aren't necessarily internal. A path on a removable volume (`/storage/XXXX-XXXX/…` — SD card, USB-OTG) also fails `exists()` **while the volume is ejected**. Sequence:

1. Library lives on SD; rows hold plain paths on the card.
2. User ejects the card, opens the app once.
3. Sweep: every external row → `!exists()` → `MISSING` → un-marked. Whole external library reset in one launch.
4. Card reinserted: files are back, but the DB rows are already cleared — Library stays empty. Permanent.

The SAF (`content://`) branch already had a safety valve for this (`INCONCLUSIVE` = do nothing on ambiguous evidence); the plain-path branch had none.

### Fix

A missing plain file now classifies `MISSING` only when its absence is trustworthy:

- path under `/data/` — internal app storage, never unmounts, `exists()` reliable (sweep's original target: junk/deleted internal downloads — behavior unchanged there), or
- the containing volume currently reports `Environment.MEDIA_MOUNTED`.

Anything else — ejected volume, unresolvable path — returns `INCONCLUSIVE`, the sweep's existing do-nothing valve. Worst case of a false `INCONCLUSIVE` is one skipped cleanup; the next launch with storage healthy re-classifies correctly and true junk still gets swept.

The mount probe (`volumeMounted`) is an injectable seam because `Environment` is Android framework — keeps the branch coverable by plain JVM tests.

### Testing

- 3 new `LocalFileOpsTest` cases: missing file on an unmounted volume → `INCONCLUSIVE`; on a mounted volume → `MISSING`; internal `/data` path never consults the mount probe.
- Full `LocalFileOpsTest` (12) + `DownloadReconcileTest` suites pass; `:core:data:compileDebugKotlin` clean.

Can't resurrect a library that was already reset — restoring those rows would need a re-sync — but after this, ejecting storage can no longer damage the DB.

### DIFF
--- a/core/data/src/main/kotlin/com/stash/core/data/files/LocalFileOps.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/files/LocalFileOps.kt
@@ -1,6 +1,7 @@
 package com.stash.core.data.files
 
 import android.content.Context
+import android.os.Environment
 import androidx.core.net.toUri
 import androidx.documentfile.provider.DocumentFile
 import com.stash.core.common.constants.StashConstants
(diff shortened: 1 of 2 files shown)


rawnaldclark:
Merged — the INCONCLUSIVE integration is exactly right, and the Windows-separator test detail was a nice catch. One non-blocking optimization if you ever revisit: the mount probe runs per file, so a large external library pays the volume resolution N times per sweep — memoizing per volume prefix would make it free. Ships in v0.9.81 within the hour.

PR #353, opened 2026-07-22
Fix duplicate playback when queueing tracks offline

### LINKED ISSUE #72: Adding two songs in queue plays the same song (offline)
Steps to replicate:
Add a song not downloaded to queue
Play it
Add another song thats not downloaded to queue
Bug happens

For offline mode, either like make users not add songs not downloaded into their library or fix the logic.

---
_Imported from Discord: https://discord.com/channels/1508402965042561054/1527913875134218310_

ParaliyzedEvo:
## Summary
- persist transient zero-ID Qobuz tracks before queueing so each track gets a distinct MediaItem and stream-cache key
- enforce Online/Offline playability at queue mutation time, including valid local file and SAF content downloads, mode-change races, and radio growth
- return queue append outcomes to UI callers and provide accurate success/failure feedback

## Testing
- [x] Focused `core:media` regression suite: 40 tests, 0 failures
- [x] `:feature:search:compileDebugKotlin`
- [x] `git diff --check`

## Known local verification limitations
- The full `:core:media:testDebugUnitTest` task timed out locally after 10 minutes without reporting a test failure; the focused affected classes pass.
- `:feature:search:testDebugUnitTest` is independently blocked on current master by a missing `sharedTrackLinkHolder` argument in `SearchViewModelTest.kt:306`.

Closes https://github.com/ParaliyzedEvo/Stash/issues/72

### DIFF
--- a/core/media/src/main/kotlin/com/stash/core/media/PlayerRepository.kt
+++ b/core/media/src/main/kotlin/com/stash/core/media/PlayerRepository.kt
@@ -145,25 +145,28 @@ interface PlayerRepository {
     /**
      * Insert [track] immediately after the currently-playing track in the queue.
      * Playback continues uninterrupted; the inserted track will play next.
+     * Returns false when the track is not playable in the current mode or the
+     * player controller is unavailable.
      */
-    suspend fun addNext(track: Track)
+    suspend fun addNext(track: Track): Boolean
 
     /**
      * Append [track] to the end of the current queue.
-     * Playback continues uninterrupted.
+     * Playback continues uninterrupted. Returns false when the track is not
+     * playable in the current mode or the player controller is unavailable.
      */
-    suspend fun addToQueue(track: Track)
+    suspend fun addToQueue(track: Track): Boolean
 
     /**
      * Append [tracks] (in order) to the end of the current queue.
      * Single MediaController.addMediaItems round-trip — preferred
      * over looping the single-track variant for known-size batches
      * like an album's full tracklist or an artist's catalog. Empty
-     * list is a no-op.
+     * list is a no-op. Returns true when at least one track was appended.
      *
      * Playback continues uninterrupted.
      */
-    suspend fun addToQueue(tracks: List<Track>)
+    suspend fun addToQueue(tracks: List<Track>): Boolean
 
     /**
      * Start a radio station seeded from an artist or track. Builds a balanced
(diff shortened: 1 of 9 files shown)


rawnaldclark:
The core fix here is valuable and I want it in — `withQueueIdentity()` persisting transient zero-ID Qobuz tracks before queueing is the right fix for the duplicate playback, and enforcing the Online/Offline playability contract at queue-mutation time is a nice hardening. It'll need a rebase before it can land, though.

Master moved since you opened this — `startRadio` was refactored from returning `Boolean` to a sealed `RadioStartResult` (Started / StreamingOff / PlayerNotReady / NoStation). Your `PlayerRepositoryImpl` conflict is right at that spot: the branch adds `if (!streamingPreference.current()) return false`, which (a) uses the old Boolean API and (b) is redundant now — `startRadio` already checks streaming and returns `RadioStartResult.StreamingOff` at the top. On rebase, that added guard can just be dropped and the surrounding code kept as `RadioStartResult`.

The bigger merge work is the tests: `PlayerRepositoryRadioTest`, `TrackActionsDelegateQueueActionsTest`, and `ListeningRecorderSkipTest` were all rewritten for the new `RadioStartResult` API, and your branch edits the same files, so those will need hand-merging (the mock's `startRadio` override now returns `RadioStartResult`, not `Boolean`).

Also good news: the `sharedTrackLinkHolder` blocker you noted in `SearchViewModelTest` is already fixed on master, so that one's off your plate. Since I own the `RadioStartResult` change that this collides with, I'm happy to do the rebase for you if you'd rather — just say the word. Otherwise a rebase on latest master + the notes above should do it.

ParaliyzedEvo:
@rawnaldclark kk, this pr should be good to go now after I rebased, lmk if I need to do anything addtional

rawnaldclark:
Rebased and merged onto master (as `4ab102a3`, Josh kept as commit author). The core fix — persisting transient zero-ID Qobuz tracks so each gets a distinct MediaItem/cache key — is the right call, and the Online/Offline playability enforcement at queue-mutation time is a nice hardening.

Two things I reconciled during the rebase, for the record:
- `startRadio` was refactored to return `RadioStartResult` on master, so the branch's post-generation streaming re-check now returns `RadioStartResult.StreamingOff` instead of `false` (kept the guard — it catches a real race where streaming toggles off during the slow generator call, and your test for it now asserts the sealed result).
- `AlbumDiscoveryViewModelTest`'s `addAlbumToQueue` test was missing the `addToQueue` Boolean stub, so it NPE'd on the suspend unbox once the tests actually ran (they didn't compile on your base due to the `sharedTrackLinkHolder` issue, which is fixed on master now). Added the stub to match your sibling tests.

Full compile-all-tests gate + the affected suites are green. Thanks!

PR #355, opened 2026-07-22
fix(sync): prevent FK failures for duplicate playlist tracks

ParaliyzedEvo:
## Summary

- coalesce repeated Spotify, YouTube, and canonical identities before the batched track insert so playlist memberships never reference IDs invalidated by Room REPLACE
- preserve distinct provider identities, normalize blank IDs, and prevent YouTube backfills from overwriting an established identity
- preserve last-occurrence position/artwork semantics and add Room-backed regressions for Spotify, YouTube Music, offline queueing, identity conflicts, and blank IDs

## Root cause

DiffWorker performed one pre-insert lookup for a whole playlist. Repeated previously unseen identities were therefore staged more than once. TrackDao.insertAll() uses OnConflictStrategy.REPLACE, so a later duplicate deleted the first generated row and left its generated ID stale. Inserting playlist_tracks with that stale ID raised SQLite foreign-key error 787 and rolled the playlist transaction back.

## Test plan

- [x] ./gradlew :core:data:testDebugUnitTest --tests com.stash.core.data.sync.workers.DiffWorkerTest --tests *TrackDaoFindExistingForBatchTest --tests *DefaultSyncEnabledTest --no-daemon
- [x] ./gradlew :app:compileDebugKotlin --no-daemon
- [x] git diff --check

## Notes

Exact repeated occurrences still collapse to one local membership because playlist_tracks is keyed by (playlist_id, track_id); the final occurrence position is retained. Supporting visible duplicate occurrences would require a schema change.

A forced full :core:data:testDebugUnitTest run currently reports five failures in unrelated existing test areas: DiscoveryQueueDaoFairnessTest, TrackDaoStreamableTest, MusicRepositoryDownloadsMixTest, and two StashDiscoveryWorkerStreamOnlyTest cases.

### DIFF
--- a/core/data/src/main/kotlin/com/stash/core/data/db/dao/TrackDao.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/db/dao/TrackDao.kt
@@ -1357,6 +1357,7 @@ interface TrackDao {
         """
         UPDATE tracks SET youtube_id = :youtubeId
         WHERE id = :trackId
+          AND (youtube_id IS NULL OR TRIM(youtube_id) = '')
           AND NOT EXISTS (
               SELECT 1 FROM tracks WHERE youtube_id = :youtubeId AND id != :trackId
           )
--- a/core/data/src/main/kotlin/com/stash/core/data/sync/workers/DiffWorker.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/sync/workers/DiffWorker.kt
@@ -89,6 +89,12 @@ class DiffWorker @AssistedInject constructor(
         private const val NEVER_MATCH_SENTINEL = "\u0000__stash_never_match__"
     }
 
+    private class BatchTrack(
+        var entity: TrackEntity,
+        val isNew: Boolean,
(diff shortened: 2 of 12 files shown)


rawnaldclark:
Thanks for jumping on this so fast — you had the right root cause and the right architecture (registering pending tracks in the per-batch identity maps) up first, and the fix that merged (#357) landed on that same approach. The merge call came down to PR mechanics only: this branch is stacked on #353, so squash-merging it would have pulled 9 unrelated media/search files in with it.

Two pieces of this PR are genuinely additive over what merged, and I'd gladly take them as a follow-up on current master:

- the strong-ID compatibility guard on canonical matches (`strongIdsCompatible`), so same-canonical tracks with conflicting provider IDs stay separate instead of merging
- the `TRIM`/unclaimed guard inside `updateYoutubeIdIfUnclaimed`

#353 stays open for review on its own merits. Sorry the two fixes crossed in flight — and thanks again for turning around a fix for this within a day of the reports.

ParaliyzedEvo:
Superseded by #357, will make a follow up pr

PR #359, opened 2026-07-22
fix(sync): reject cross-identity merges when canonical match carries a conflicting provider ID

ParaliyzedEvo:
## Summary
- Follow-up to #357 — restores the `strongIdsCompatible` guard from #355 that didn't make it into the merged fix, so a canonical-title/artist match is only reused when neither snapshot's spotifyUri/youtubeId actively conflicts with what's already stored on that row.
- Without the guard, two distinct YouTube uploads of the same song (same canonical title/artist, different videoIds) collapse into a single track row instead of staying separate — the second identity's videoId is silently dropped.
- `byCanonical` changes from a single-value map to a list per canonical key, since a canonical key can now legitimately resolve to more than one row.

## Root cause
`DiffWorker`'s bulk identity resolution matches a new/incoming snapshot against in-batch and in-DB candidates in priority order: spotifyUri → youtubeId → canonical identity. The canonical fallback in master takes whatever `BatchTrack` last registered under that (title, artist) key with no compatibility check — so if two snapshots share a canonical key but carry different youtubeIds, the second snapshot's canonical lookup returns the first snapshot's row and merges into it, discarding its own youtubeId.

## Changes
- `core/data/.../sync/workers/DiffWorker.kt`
  - `byCanonical` is now `MutableMap<CanonicalKey, MutableList<BatchTrack>>`
    instead of a single-value map.
  - `registerBatchTrack` appends to the per-key list instead of
    overwriting it.
  - New `strongIdsCompatible(track, snapshot)` — true only when neither
    side's spotifyUri/youtubeId is non-blank AND disagrees with the other.
  - `matchBatchTrack`'s canonical fallback now filters candidates through
    `strongIdsCompatible` via `firstOrNull { ... }` instead of returning
    an unconditional single match.
- `core/data/.../sync/workers/DiffWorkerTest.kt`
  - Added `distinct YouTube IDs with the same canonical metadata stay separate`
  - Added `existing canonical match reserves its first YouTube identity`

## Test plan
- [x] `./gradlew :core:data:testDebugUnitTest --tests com.stash.core.data.sync.workers.DiffWorkerTest` — full suite passes, including the two new regression tests

### DIFF
--- a/core/data/src/main/kotlin/com/stash/core/data/sync/workers/DiffWorker.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/sync/workers/DiffWorker.kt
@@ -399,20 +399,42 @@ class DiffWorker @AssistedInject constructor(
         }
         val bySpotifyUri = mutableMapOf<String, BatchTrack>()
         val byYoutubeId = mutableMapOf<String, BatchTrack>()
-        val byCanonical = mutableMapOf<CanonicalKey, BatchTrack>()
+        // A canonical key can legitimately map to MULTIPLE rows — e.g. two
+        // distinct YouTube uploads of "the same song" that arrive with
+        // different videoIds. A single-value map here would collapse them
+        // into one row, silently discarding the second identity.
+        val byCanonical = mutableMapOf<CanonicalKey, MutableList<BatchTrack>>()
 
         fun registerBatchTrack(track: BatchTrack) {
             track.entity.spotifyUri?.takeIf(String::isNotBlank)?.let { bySpotifyUri[it] = track }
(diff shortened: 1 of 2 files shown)


rawnaldclark:
Merged — thanks for tracking this one down. Restoring the `strongIdsCompatible` guard that got dropped from #355 is exactly right, and the `byCanonical` list change is the correct fix for the collapse: two YouTube uploads sharing a canonical key legitimately need to stay separate rows. The two regression tests (`distinct YouTube IDs…` / `existing canonical match reserves…`) are what made this an easy merge — they pin the behavior so it can't silently regress again.

Verified it compiles and the full `DiffWorkerTest` suite passes against current master (which has moved a fair bit since you opened this) before merging.

PR #356, opened 2026-07-22
UI improvements: playlist pinning, view album, sleep timer, and AMOLED polish

ParaliyzedEvo:
## Summary
- **Playlist pinning** — users can now pin favorite playlists to the top of the Library Playlists grid via a new context-menu action. Backed by a new `pinned` column on the `playlists` table (Room migration v34→v35), with pinned playlists sorted alphabetically ahead of the rest.
- **"View Album" from track context menu** — tapping a track's ⋮ menu now offers a "View Album" action that resolves the track's album via the YouTube Music API and navigates to the album detail screen, rather than only being reachable from the local Albums tab.
- **Now Playing sleep timer improvements** — the sleep timer gained expanded controls in `NowPlayingScreen` (custom duration entry, "end of track" mode) backed by new playback-service and view-model logic for scheduling/cancelling the timer.
- **Now Playing screen redesign** — broader layout rework of `NowPlayingScreen` for a cleaner visual hierarchy.
- **AMOLED blur option** — added a new theme preference to toggle background blur under the AMOLED theme, threaded through `ThemePreference`, `ThemePreferencesManager`, `SettingsViewModel`, and the ambient background / live lyrics bar composables.
- **Save to Playlist from track context menu** — added a "Save to Playlist" action directly in the track's BottomSheet menu, saving a navigation hop.
- **Qobuz discovery status banner** — added a banner shown on the Home screen when the Qobuz discovery toggle is on but there's no internet connection or no token configured, so the feature's disabled state is no longer silent. Follow-up commit clarified the banner's copy with setup instructions.
- **Scrollbar hitbox tuning** — lowered the horizontal hitbox on the draggable fade scrollbar (`ScrollbarModifier.kt`) to reduce accidental horizontal-scroll conflicts.

### Playlist pinning
- `core/data/.../db/StashDatabase.kt` — bumped schema to v35, added `pinned` column + `MIGRATION_34_35`.
- `core/data/.../db/dao/PlaylistDao.kt`, `entity/PlaylistEntity.kt`, `mapper/PlaylistMapper.kt` — plumbed the new column through the DAO/entity/mapper.
- `core/data/.../repository/MusicRepository.kt` + `MusicRepositoryImpl.kt` — added `setPlaylistPinned(...)`.
- `core/model/Playlist.kt` — added `pinned: Boolean` to the domain model.
- `feature/library/.../LibraryScreen.kt` — pin/unpin action in the playlist context-menu bottom sheet, pinned indicator icon on grid items.
[...]

PR #362, opened 2026-07-24
Implement functional search field in Settings hub

ParaliyzedEvo:
## Summary
- Replaces the non-functional Settings search pill (Phase 1 placeholder — static text, no input) with a real text field that filters across the six settings categories by title and keyword, navigating to the matching category on tap.
- Does not deep-link to individual rows within a category — landing on the right category screen is the intended scope for this pass.

## Changes
- `feature/settings/.../components/SettingsSearchField.kt`
  - Rewritten from a static clickable `Row` to a hoisted-state `BasicTextField` (`value`/`onValueChange`), with a clear ("×") affordance when non-empty.
- `feature/settings/.../SettingsHubScreen.kt`
  - New local `searchQuery` state and a `searchIndex` of the six categories (title, subtitle, keywords, icon, nav callback).
  - New `searchResults`: case-insensitive filter over title + keywords.
  - While `searchQuery` is non-blank, renders matching categories (or a "no results" message) in place of the default six-row card; tapping a result clears the query and navigates via the existing `onOpen*` callback.

## Test plan

### DIFF
--- a/core/media/src/main/kotlin/com/stash/core/media/PlayerRepository.kt
+++ b/core/media/src/main/kotlin/com/stash/core/media/PlayerRepository.kt
@@ -104,6 +104,12 @@ interface PlayerRepository {
     /** Pause playback. */
     suspend fun pause()
 
+    /**
+     * Set output volume, 0f (silent) to 1f (full). Used by
+     * [com.stash.core.media.SleepTimerController] to fade out before pausing.
+     */
+    fun setVolume(volume: Float)
+
     /** Skip to the next track in the queue. */
     suspend fun skipNext()
 
--- a/core/media/src/main/kotlin/com/stash/core/media/PlayerRepositoryImpl.kt
+++ b/core/media/src/main/kotlin/com/stash/core/media/PlayerRepositoryImpl.kt
@@ -364,6 +364,18 @@ class PlayerRepositoryImpl @Inject constructor(
         ensureController()?.pause()
     }
 
+    override fun setVolume(volume: Float) {
+        // Not suspend — SleepTimerController's fade loop calls this every
+        // ~350ms and can't afford a full ensureController() connect-await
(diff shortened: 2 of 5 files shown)


rawnaldclark:
The Settings-hub search feature itself looks good and is independent — happy to merge that part. One thing to sort out first: this branch also carries two commits that add `setVolume(...)` to the `PlayerRepository` interface + impl + the test mock (`Add setVolume method to PlayerRepository…` / `…to mock…`). That's the same `setVolume` your sleep-timer work in #356 adds — so as-is, #362 and #356 both declare it and would collide, and here on current master the `setVolume` would land as an unused interface method (nothing calls it until #356's `SleepTimerController` lands).

Could you drop those two `setVolume` commits from this branch, leaving just `Implement functional search field in Settings hub`? Then it merges cleanly and standalone, and `setVolume` stays owned by #356 where it's actually used. Heads-up while you're at it: master now runs a compile-all-modules unit-test gate in CI, so a mock referencing an interface method that isn't present will fail the build — worth a rebase on latest master to be safe.

rawnaldclark:
Landed the search feature on master 🎉 — cherry-picked your `Implement functional search field in Settings hub` commit directly (`748ff9da`), so your authorship is preserved on the commit. It applied cleanly and passed the full compile-all-modules test gate.

I split it out rather than merging the whole PR because the branch also carried the two `setVolume` commits, which belong with the sleep-timer work in #356 (where `SleepTimerController` actually calls it) — landing them here would've put an unused interface method on master and collided with #356. So `setVolume` stays yours over in #356.

Closing this since the feature's merged. One small future note (not blocking, already shipped): the `searchIndex` is a third hardcoded copy of the six categories (alongside `settingsHubSummaries` and the default rows), so if a 7th settings category ever gets added, it'll need adding to the index too or search won't find it — worth folding into a single source down the line. Really nice, clean feature. Thanks!

PR #364, opened 2026-07-24
fix(sync): stop manual re-sync from cancelling in-flight sync + show live fetch/diff progress

ParaliyzedEvo:
## Summary
- Fixes a bug where tapping "Sync Now" while a sync was already running would cancel the in-flight `PlaylistFetchWorker` chain outright (`ExistingWorkPolicy.REPLACE`) and restart from zero — every partially-completed playlist/mix fetch threw `JobCancellationException`, producing repeated failed syncs on large libraries (log evidence: 6 manual re-triggers in ~10 minutes, all ending in `WorkerStoppedException`).
- Root cause was compounded by poor progress visibility: `FetchingPlaylists` and `Diffing` were flat progress markers (20%/25%) with no live counter, so on a large library (4k+ tracks) the bar could sit visually frozen for 40+ minutes before any download began — driving users to repeatedly tap Sync Now, believing it had stalled. This is the same root cause reported in a user-filed enhancement request proposing separate remote-sync vs. local-verification progress.
- This PR fixes the re-trigger bug and adds live progress counters for both phases. It does not yet split fetch/diff into a standalone "Library Verification" action outside the sync chain — that's scoped as a follow-up.

## Changes
- `core/data/.../sync/SyncScheduler.kt`
  - `triggerManualSync()` now checks `isSyncChainActive()` (queries `WorkManager.getWorkInfosForUniqueWork` for any ENQUEUED/RUNNING state) and no-ops if a sync is already running, instead of unconditionally enqueuing with `REPLACE`.
- `feature/sync/.../SyncViewModel.kt`
  - `onSyncNow()` also short-circuits when `uiState.isSyncing` is true, so the redundant call never reaches the scheduler.
- `core/data/.../sync/SyncPhase.kt`
  - `FetchingPlaylists` is now a data class carrying `playlistsFetched: Int` (live count; no fixed total is knowable upfront for either source).
  - `Diffing` is now a data class carrying `playlistsDiffed`/`totalPlaylists: Int` (real fraction — the diff worker knows its total upfront).
- `core/data/.../sync/SyncStateManager.kt`
  - `onFetchingPlaylists()`/`onDiffing()` now take progress params and can be called repeatedly to update the live phase state.
- `core/data/.../sync/workers/PlaylistFetchWorker.kt`
  - Added a thread-safe `AtomicInteger` counter, incremented after every playlist/mix/liked-songs fetch (Spotify daily mixes, liked songs, user playlists; YouTube home mixes, liked songs, user playlists), reporting through `syncStateManager.onFetchingPlaylists(count)`.
- `core/data/.../sync/workers/DiffWorker.kt`
  - Reports `syncStateManager.onDiffing(diffed, total)` before the loop starts and after each playlist is reconciled.
- `feature/sync/.../components/SyncActionProgress.kt`
  - `phaseLabel()` now renders the live counts: "Fetched N playlists..." during fetch, "Comparing changes (N/total)..." during diff, instead of static unchanging text.

## Test plan

### DIFF
--- a/core/data/src/main/kotlin/com/stash/core/data/sync/SyncPhase.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/sync/SyncPhase.kt
@@ -25,14 +25,44 @@ sealed interface SyncPhase {
         override val progress: Float = 0.05f
     }
 
-    /** Retrieving playlist metadata from the remote source. */
-    data object FetchingPlaylists : SyncPhase {
-        override val progress: Float = 0.20f
+    /**
+     * Retrieving playlist metadata from the remote source.
+     *
+     * @property playlistsFetched Running count of playlists/mixes fetched so
+     *   far this phase. No total denominator: the true playlist count for
+     *   Spotify (folder-walked) and YouTube isn't known until the fetch
+     *   itself finishes, so this is a live counter for display ("Fetched 42
+     *   playlists…") rather than a fraction — a fabricated total would be
+     *   worse than none. [progress] still nudges forward (capped) as the
+     *   count grows so a long fetch visibly moves instead of sitting dead
+     *   at the phase's starting value.
+     */
+    data class FetchingPlaylists(
+        val playlistsFetched: Int = 0,
+    ) : SyncPhase {
+        override val progress: Float
+            get() = (0.05f + minOf(playlistsFetched, 150) * 0.001f).coerceAtMost(0.20f)
     }
 
-    /** Comparing remote snapshots against local data to find differences. */
-    data object Diffing : SyncPhase {
-        override val progress: Float = 0.25f
+    /**
+     * Comparing remote snapshots against local data to find differences.
+     *
+     * @property playlistsDiffed Playlists reconciled so far this phase.
+     * @property totalPlaylists  Total playlists to reconcile — known
+     *   upfront here (unlike fetch), since the snapshot rows already exist
+     *   in the DB by the time this phase starts.
+     */
+    data class Diffing(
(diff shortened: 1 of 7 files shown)


rawnaldclark:
Merged — this is a great root-cause fix. The insight that impatient re-taps were *causing* the failures (REPLACE cancelling the in-flight chain, so the re-tapping was what kept sync from ever finishing) is spot on, and pairing the guard with live progress counters addresses *why* users re-tapped in the first place rather than just suppressing the symptom. It also complements a `CancellationException`-rethrow fix that recently landed in `DiffWorker`/`SyncFinalizeWorker` — together they clean up cancelled-vs-failed sync reporting from both ends.

Verified the compile gate + `DiffWorkerTest` pass on current master before merging.

One follow-up nit (didn't block the merge): `isSyncChainActive()` calls `workManager.getWorkInfosForUniqueWork(...).get()`, which blocks the calling thread. The comment says it runs off-main via `viewModelScope`, but `onSyncNow()` isn't a coroutine — it calls `triggerManualSync()` synchronously on the main thread (Compose onClick). It's gated behind the new `isSyncing` short-circuit and it's only a fast local WorkManager query, so real-world impact is small, but a synchronous `.get()` on main is a latent jank source. Worth wrapping in `withContext(Dispatchers.IO)` (or launching `onSyncNow`'s scheduler call in a coroutine) in a future pass.

Separately: this fixes the manual re-trigger REPLACE path. There's a sibling case — `scheduleDailySync` also uses REPLACE, so editing sync settings while a sync runs can cancel it — if you ever want to pick that up as a follow-up.

PR #384, opened 2026-07-30
Fix multiple issues with syncing songs

### LINKED ISSUE #343: [Bug]: Sync Errors leading to reduced songs instead of adding more
### Before you start

### Stash version

v0.9.75 all throughout 0.9.81

### Phone & Android version

Pocophone F1 (Pixel 8 OS), Android 14

### Where does it happen?

Actual Result: Syncing leads to errors and deleted songs instead of adding more.

Expected Result: Syncing should add more songs based on what playlists are selected.

### Steps to reproduce

1. Select relevant playlists from the sync screen (both YT and Spotify sections).
2. Click on Download tracks to device (app is in offline mode)
3. Observe

Note: This is slightly different from bug 337

Video at https://anonmp4.art/v/14Dil3T38deaPso

Image
Image
Image
Image
Image
Image
Image
Image
Image
Image
Image

ParaliyzedEvo:
## Summary
- `PlaylistFetchWorker` now tracks a run-wide `youtubeInventoryComplete: AtomicBoolean`, flipped to `false` by `markYoutubeIncomplete(reason)` the moment any leg of the YouTube fetch during that sync run fails or throws (home mixes, liked songs, user playlist listing, or any individual mix/playlist track fetch). The final state is exported through `KEY_YOUTUBE_INVENTORY_COMPLETE` in the worker's output data.
- `DiffWorker` reads that flag and only calls `PlaylistDao.deactivateMissingForSource(YOUTUBE, ...)` when the inventory was reported complete. If the fetch was partial, it logs `"Skipping stale-YouTube-playlist deactivation — this sync's inventory was incomplete"` and leaves all existing YouTube playlists untouched. Previously, a single flaky sub-fetch (such as one mix's tracks failing or one page of the playlist listing erroring) made every other YouTube playlist appear "missing" for that sync and caused them to be soft-hidden from Home even though they still existed.
- This mirrors the equivalent Spotify-side safeguard introduced in #343 (`shouldDeactivateMissingSpotifyPlaylists` / partial playlist snapshot flag) by applying the same protection to YouTube playlist syncs.
- Includes a small follow-up fix for a compile break introduced by the inventory-completeness change in `PlaylistFetchWorker.kt`, resolving a return-type/call-site mismatch around the liked-songs fetch boolean feeding into `youtubeInventoryComplete`.
- `PlayerRepositoryImpl`'s `onMediaItemTransition` now defensively calls `controller.prepare()` whenever the player lands in `STATE_IDLE` while a current media item is still set. Some failure paths (such as audio offload sink stalls or certain codec/format edge cases) could leave the player idle on the next track without firing `onPlayerError`, causing auto-advance to display the next track in Now Playing while the Play button did nothing. A no-op `prepare()` when already ready safely recovers this state automatically.
- `CrossfadeEngine`'s manual audio focus listener now correctly handles transient audio focus loss (`AUDIOFOCUS_LOSS_TRANSIENT`, such as navigation prompts) that occurs during an active crossfade. Previously, only the outgoing/master player's `playWhenReady` state was tracked for resume on `AUDIOFOCUS_GAIN`, so the spare player (already audible and about to become the master) could remain paused after focus returned, resulting in silence at the transition despite playback appearing active. The gain handler now also resumes the spare player when a transition is in progress.
- `performTransition` now checks `isActive` before executing its post-fade tail (queue transfer, role swap, and cleanup). This ensures a `cancelTransition()` arriving after the fade loop's final `delay()`—but before the swap completes—actually aborts the transition instead of promoting the wrong player and wiring it to the media session.
- `StashPlaybackService` was updated to wire these changes through its `onPlayerError` and `onMediaItemTransition` listener paths so crossfade cancellation and transient audio focus recovery remain consistent with master/spare player swaps.

## Changes
- `core/data/src/main/kotlin/com/stash/core/data/sync/workers/PlaylistFetchWorker.kt`
  - Added `youtubeInventoryComplete: AtomicBoolean` and `markYoutubeIncomplete(reason)`, invoked from every YouTube fetch failure path (`getHomeMixes`, `getUserPlaylists`, per-mix/per-playlist track fetch, and `getLikedSongs`)
  - `fetchYouTubePlaylists` now resets the flag at the start of each run and returns the final completion state
  - `fetchAndSnapshotLikedSongs` now returns `Boolean` so liked-song failures also mark the inventory as incomplete
  - `doWork` now includes `KEY_YOUTUBE_INVENTORY_COMPLETE` in the success `workDataOf`
  - Fixed the compile error caused by the initial change by resolving the liked-songs return-type/call-site mismatch
- `core/data/src/main/kotlin/com/stash/core/data/sync/workers/DiffWorker.kt`
  - Reads `KEY_YOUTUBE_INVENTORY_COMPLETE` (defaulting to `false` for fail-closed behavior)
  - Gates `deactivateMissingForSource(YOUTUBE, ...)` on a complete inventory and logs a skip message when the inventory was incomplete
- `core/media/src/main/kotlin/com/stash/core/media/PlayerRepositoryImpl.kt`
  - Added a defensive `controller.prepare()` call in `onMediaItemTransition` when playback reaches `STATE_IDLE` with a current media item still present
- `core/media/src/main/kotlin/com/stash/core/media/service/CrossfadeEngine.kt`
  - Updated transient audio focus recovery so the spare player is also resumed when focus returns during an active crossfade
  - Added an `isActive` guard before the post-fade transition tail in `performTransition` so cancelled transitions cannot complete the player swap
- `core/media/src/main/kotlin/com/stash/core/media/service/StashPlaybackService.kt`
  - Updated listener wiring to keep `onPlayerError`/`onMediaItemTransition` behavior consistent with the new crossfade cancellation and audio focus resume logic

## Test plan
- [x] `./gradlew test` (affected modules)

Closes: #343

### DIFF
--- a/core/data/src/main/kotlin/com/stash/core/data/sync/workers/DiffWorker.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/sync/workers/DiffWorker.kt
@@ -120,6 +120,12 @@ class DiffWorker @AssistedInject constructor(
 
     override suspend fun doWork(): Result {
         val syncId = inputData.getLong(PlaylistFetchWorker.KEY_SYNC_ID, -1L)
+        // Defaults to false (fail-closed) — if the key is somehow missing, treat
+        // the inventory as unreliable and skip deactivation rather than risk
+        // hiding real playlists.
+        val youtubeInventoryComplete = inputData.getBoolean(
+            PlaylistFetchWorker.KEY_YOUTUBE_INVENTORY_COMPLETE, false,
+        )
         if (syncId == -1L) {
             syncStateManager.onError("DiffWorker: missing sync ID")
             return Result.failure()
@@ -216,7 +222,11 @@ class DiffWorker @AssistedInject constructor(
             val youtubeSourceIds = playlistSnapshots
                 .filter { it.source == MusicSource.YOUTUBE }
                 .map { it.sourcePlaylistId }
-            if (youtubeSourceIds.isNotEmpty()) {
+            // #post-343: mirrors the Spotify inventoryComplete guard — never treat a
+            // PARTIAL YouTube fetch (one failed sub-leg: home mixes / liked songs /
+            // a specific playlist's tracks) as "these other playlists are gone."
+            // A single flaky call used to hide the user's entire YouTube library.
+            if (youtubeSourceIds.isNotEmpty() && youtubeInventoryComplete) {
                 val hidden = playlistDao.deactivateMissingForSource(
                     source = MusicSource.YOUTUBE,
                     currentSourceIds = youtubeSourceIds,
@@ -224,6 +234,8 @@ class DiffWorker @AssistedInject constructor(
                 if (hidden > 0) {
                     Log.i(TAG, "Deactivated $hidden stale YouTube playlist(s)")
                 }
+            } else if (youtubeSourceIds.isNotEmpty()) {
+                Log.w(TAG, "Skipping stale-YouTube-playlist deactivation — this sync's inventory was incomplete")
             }
 
             // Clean up orphaned tracks whose playlists were refreshed and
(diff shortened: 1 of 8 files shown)


rawnaldclark (commented):
Disregard my earlier notes here — sorry for the noise. Both fixes are good and I'm merging as-is; I'll take care of the small follow-ups (restoring the #326 crossfade identity log and the requestFocus rationale comment) on master myself rather than sending this back to you. Thanks for the sync fix especially — gating deactivation on inventory completeness is the right shape.

PR #390, opened 2026-07-31
ListenBrainz scrobbling, and mirror hearts to Last.fm loved tracks

rawnaldclark:
Two long-requested integrations, plus the refactor that made the second one cheap.

## ListenBrainz scrobbling

MetaBrainz's open scrobbling service. Token auth (paste from `listenbrainz.org/profile`), validated before it's stored, with a now-playing toggle and manual sync.

**Verified end to end on device** — three listens landed on a real profile:

```
17:04:15  Liz Phair — Glory     [Exile In Guyville]  via Stash
16:56:41  Overmono — So U Kno   [Good Lies]          via Stash
16:55:11  Liz Phair — Glory     [Exile In Guyville]  via Stash
```

Baseline was 0, so all three are ours. The two older ones matter most: they were submitted *before* the account's email was verified, got 401'd, sat queued, and **submitted themselves once the account was fixed** — with no user action. That's the retry design proving itself.

## Last.fm loved tracks (#—)

Stash already *read* `user.getLovedTracks` to seed Liked Songs and mirrored hearts to Spotify and YouTube, but never called `track.love`. So hearting a track appeared everywhere except the service whose loves you'd imported.

Last.fm joins as a third `Destination` in the existing `LikeCoordinator` queue, inheriting its pacing, fire-time pref re-read, and organic retry for free.

One way it differs, and it needed its own test: **Last.fm matches on artist + title**, so there's no platform id to resolve and no id precondition. Spotify and YT skip a track with no `spotifyUri`/`youtubeId`; reusing that "add if id present" shape would have silently dropped every love for exactly the tracks most likely to need one.

`tracks.lastfm_loved_at` (v38) is **not** deduplication bookkeeping — `track.unlove` destroys data Stash doesn't own. Someone may have loved a track on Last.fm years before installing this app, and un-hearting here must not delete that. Only a love this column records is ever un-loved.

## The refactor underneath

`LastFmScrobbler` and `YouTubeHistoryScrobbler` each carried their own copy of "load pending, resolve track, submit, mark done, retry later" — the second says so in its own KDoc. ListenBrainz would have been the third copy, plus a third boolean column and a third DAO triplet (`scrobbled`, then `yt_scrobbled`, then…).

`listen_submissions` (v37) makes a destination a `target` string instead. Only outcomes are stored, so pending work is "no SENT row for this target" — same semantics, no bookkeeping.

Two failure modes are distinguished deliberately, because conflating them loses data either way:

- **Transient** (5xx, timeout, a sink that throws, **and 401**): held with attempt counts untouched. A service down for a week must never cost a week of history.
- **Rejected** (other 4xx): batch is split and each listen submitted alone, so one malformed row can't burn its neighbours' retries.

401 moved to Transient *because of* the device test above — it's an account problem, not a listen problem, and classifying it Rejected would have discarded perfectly good listens at the attempt cap.

## Guards worth calling out

**New destinations start from now.** Stash keeps every play it has ever recorded; without a cutoff, connecting would submit thousands of old listens as one flood nobody asked for. The sink **fails closed** — unknown cutoff returns `Long.MAX_VALUE` and submits nothing rather than defaulting to zero. Reconnecting the same token deliberately does *not* move the cutoff forward, which would discard anything queued during the outage you're reconnecting to fix.

**Only ListenBrainz is on the generic path.** Last.fm and YouTube history keep their own scrobblers and columns until this proves itself in the wild.

## Testing

1936 tests green repo-wide. Both schema exports committed (37.json, 38.json).

Two bugs the device test caught that unit tests had passed:

1. **`validate-token` answers HTTP 200 for a bad token** (`{"valid":false}`) — connecting with the literal string `not-a-real-token-12345` succeeded. My test asserted `401 → false`, a response the endpoint never gives, so it was green while the real path was broken.
2. The 401 classification above.

Both now tested against shapes verified with curl first.


rawnaldclark (commented):
Self-review note for the record: `ListenSinkCoordinator` currently registers only ListenBrainz. Last.fm and YouTube history still run their own scrobblers against their own boolean columns (`scrobbled`, `yt_scrobbled`), which was the deliberate order — prove the generic path on the new destination first. Tracking so it doesn't quietly become three permanent mechanisms instead of two.

PR #391, opened 2026-07-31
feat(nowplaying): keep screen on in lyrics sheet

### LINKED ISSUE #382: [Feature]: keep awake when viewing lyrics
### Before you start

- [x] I [searched existing issues](https://github.com/rawnaldclark/Stash/issues?q=is%3Aissue) and nobody's already requested this

### What problem does this solve?

Phone falls asleep so I have to keep waking it

### What would you like Stash to do?

Stay awake

### Anything else? (optional)

_No response_

chaudhary-lakshay:
Closes #382.

### Problem

The lyrics sheet is a read-along surface, but the display timeout still fires, so the phone sleeps mid-song and has to be woken repeatedly.

Nothing in the repo currently touches keep-screen-on — `keepScreenOn` / `FLAG_KEEP_SCREEN_ON` had zero matches before this change.

### Change

Six lines in `LyricsBottomSheet.kt`:

```kotlin
val view = LocalView.current
DisposableEffect(view) {
    val wasKeepingScreenOn = view.keepScreenOn
    view.keepScreenOn = true
    onDispose { view.keepScreenOn = wasKeepingScreenOn }
}
```

Two deliberate details:

- **Scope.** `NowPlayingScreen.kt:265` guards the composable with `if (showLyrics)`, so the composable exists exactly while the sheet is open. The flag goes up on open and comes down on dismiss — no separate lifecycle to manage, and nothing is held while the sheet is closed.
- **Restore, don't force false.** `onDispose` puts back the previous value instead of hard-setting `false`, so this can never clear a keep-screen-on flag set by something else on the same view.

No setting was added. The sheet is an explicit foreground action that the user dismisses, and Android's guidance is to hold the flag for as short a time as possible — which this already does. Happy to put it behind a toggle if you'd prefer.

### Scope

`LiveLyricsBar` (the always-visible one-line ticker) is intentionally untouched — keeping the screen on for a persistent bar would be a very different battery tradeoff.

### Verification

- `./gradlew :feature:nowplaying:compileDebugKotlin` — BUILD SUCCESSFUL, no new warnings from this file.
- **Not exercised on a device.** The lifetime argument above is from reading the call site, not from watching the screen stay lit. Worth a quick manual check before merge.

### DIFF
--- a/feature/nowplaying/src/main/kotlin/com/stash/feature/nowplaying/ui/LyricsBottomSheet.kt
+++ b/feature/nowplaying/src/main/kotlin/com/stash/feature/nowplaying/ui/LyricsBottomSheet.kt
@@ -24,8 +24,10 @@ import androidx.compose.material3.Text
 import androidx.compose.material3.TextButton
 import androidx.compose.material3.rememberModalBottomSheetState
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalView
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.unit.dp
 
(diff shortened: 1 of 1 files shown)


rawnaldclark (commented):
Nice, tight change. Approving — a couple of things I checked rather than assumed:

**The comment's claim holds.** It says the composable "only exists while the sheet is open (the call site guards it with `if (showLyrics)`)", and that's the load-bearing assumption — if it were wrong, the screen would never sleep again. Verified: `NowPlayingScreen.kt:265` is exactly `if (showLyrics) {`, single call site, so `onDispose` genuinely fires on close.

**Save-and-restore is the right instinct.** Restoring `wasKeepingScreenOn` rather than forcing `false` is what I'd want. It's currently belt-and-braces — there is no other `keepScreenOn` or `FLAG_KEEP_SCREEN_ON` anywhere in the app, so it will always restore to false today — but it means whoever adds one later doesn't get silently clobbered by this.

**`LocalView.current` is read outside the `ModalBottomSheet`**, so it resolves to the host view rather than the sheet's dialog window. That's the correct one for this; inside, it might not affect the main window at all. Worth noting since it's easy to "tidy" that line into the sheet body later and quietly break it.

One thought, not a blocker: the flag is tied to *sheet visibility*, not *playback*. Open lyrics, let the track end, put the phone down — the screen stays on indefinitely. Tying it to "lyrics open **and** playing" would match intent more closely. Happy to merge as-is and treat that as a follow-up if it comes up.

PR #392, opened 2026-07-31
feat(lastfm): add first-artist-only scrobble toggle

### LINKED ISSUE #112: Only use first artist on scrobble
Hi! Could you please add an option to scrobble only the primary/first artist of a track? Currently, collabs (like "Artist A & Artist B") create separate entries on Last.fm, which clutters the library. An "Only use first artist" toggle would be amazing to keep our stats organized.

chaudhary-lakshay:
Closes #112.

### Problem

Collabs scrobble as `Artist A, Artist B`. Last.fm treats that whole string as a distinct artist from either member, so collabs fragment the user's stats instead of crediting the primary artist.

### Change

New opt-in toggle under Settings → Accounts → Last.fm (visible only when connected), off by default. When on, `LastFmScrobbler` submits only the primary artist — for both `track.scrobble` and `track.updateNowPlaying`.

```kotlin
internal fun primaryArtist(artist: String): String =
    artist.substringBefore(", ").trim().ifBlank { artist }
```

### Why split on `", "` and nothing else

Every parser in the app joins multiple artists with exactly `", "`:

- `PlaylistFetchWorker.kt:410,518,672` — `track.artists.joinToString(", ") { it.name }`
- `ytmusic/AlbumResponseParser.kt:126`
- `ytmusic/ResponseParserHelpers.kt:110`
- `ytmusic/SearchResponseParser.kt:140`

So the join delimiter is the split delimiter. I deliberately did **not** write a regex over `&` / `feat.` / `ft.` / `x` / `vs.` — that shreds real band names (`Simon & Garfunkel`, `Earth, Wind & Fire`, `Florence + the Machine`). A covered case in the tests.

### Known limitation — please read before merging

A single artist whose own name contains `", "` gets truncated: **`Tyler, The Creator` → `Tyler`**. Same for `Crosby, Stills & Nash`.

There is no way to distinguish that from a two-artist join at this layer, because by the time the string reaches the scrobbler the structure is already gone. A correct fix means carrying a structured `List<String>` of artists on `TrackEntity` — a schema change, a migration, and touching every parser. That felt well out of scope for this issue, so instead:

- the setting is **opt-in and off by default**
- the limitation is stated in the toggle's own subtitle, so a user opting in sees it
- it's documented in KDoc on `primaryArtist`

If you'd rather have the structured-artist version, say so and I'll close this — it's a much bigger change but it would be exactly right rather than approximately right.

### Deliberately out of scope

**ListenBrainz is untouched.** It has its own submission path (`ListenBrainzApiClient.kt:141` sends `listen.artist` as `artist_name`), so enabling this leaves ListenBrainz scrobbles unchanged. That's an inconsistency, but the issue asked for Last.fm and I didn't want to make a product decision for a second service unasked. Happy to extend it if you want one shared "submission artist formatting" preference.

**No backfill.** Already-submitted scrobbles are left alone.

### Notes on the implementation

- The preference lives in the existing `lastfm_session` DataStore alongside the session. It is deliberately **not** cleared by `clear()`, so disconnecting and reconnecting preserves the user's choice.
- `SettingsViewModel` exposes it as its own `StateFlow` rather than folding it into the main `combine`, which is index-based (`values[8]`, `values[9]`…) and would have required shifting every downstream index.
- The toggle row is hand-rolled rather than reusing `SettingsToggleRow`, because this section renders inside a `GlassCard` that already applies 16.dp padding and `SettingsToggleRow` adds its own 15.dp — they'd double up and misalign against the surrounding text.

### Verification

- `./gradlew :feature:settings:compileDebugKotlin :core:data:testDebugUnitTest --tests "*PrimaryArtistTest"` — BUILD SUCCESSFUL
- `PrimaryArtistTest` — 6 tests, 0 failures (single artist, comma-joined pair, `&` band name, `feat.` suffix, leading-comma fallback, blank input)
- **Not exercised on a device.** The toggle has not been tapped in a running app and no real scrobble has been submitted with it on. Worth a manual pass before merge.

### DIFF
--- a/core/data/src/main/kotlin/com/stash/core/data/lastfm/LastFmScrobbler.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/lastfm/LastFmScrobbler.kt
@@ -86,10 +86,11 @@ class LastFmScrobbler @Inject constructor(
     suspend fun notifyNowPlaying(artist: String, track: String, album: String? = null) {
         val session = runCatching { sessionPreference.session.firstOrNull() }.getOrNull()
             ?: return
+        val submitArtist = scrobbleArtist(artist)
         runCatching {
             apiClient.updateNowPlaying(
                 sessionKey = session.sessionKey,
-                artist = artist,
+                artist = submitArtist,
                 track = track,
                 album = album,
             )
@@ -128,7 +129,7 @@ class LastFmScrobbler @Inject constructor(
     private suspend fun submit(session: LastFmSession, event: ListeningEventEntity, track: TrackEntity) {
         val result = apiClient.scrobble(
             sessionKey = session.sessionKey,
-            artist = track.artist,
+            artist = scrobbleArtist(track.artist),
             track = track.title,
             album = track.album.takeIf { it.isNotBlank() },
             timestampEpochSeconds = event.startedAt / 1000,
@@ -141,7 +142,29 @@ class LastFmScrobbler @Inject constructor(
         }
     }
 
+    /**
+     * Applies the "only first artist" preference, if the user enabled it.
+     */
+    private suspend fun scrobbleArtist(artist: String): String =
+        if (sessionPreference.firstArtistOnly.firstOrNull() == true) primaryArtist(artist) else artist
+
     companion object {
         private const val TAG = "LastFmScrobbler"
     }
 }
+
+/**
+ * Best-effort primary artist for scrobbling. Every parser in the app joins
+ * multiple artists with `", "` (PlaylistFetchWorker, ResponseParserHelpers,
+ * SearchResponseParser), so the join delimiter is the split delimiter.
+ *
+ * Deliberately NOT a regex on "&" / "feat." / "x" — that shreds real band
+ * names like "Simon & Garfunkel".
+ *
+ * Known limitation: a single artist whose own name contains ", " is
+ * truncated ("Tyler, The Creator" -> "Tyler"). Fixing that properly needs a
+ * structured artist list on TrackEntity and a DB migration. This is why the
+ * setting is opt-in and off by default.
+ */
+internal fun primaryArtist(artist: String): String =
+    artist.substringBefore(", ").trim().ifBlank { artist }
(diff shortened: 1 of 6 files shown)


rawnaldclark (commented):
This is good work, and the reasoning in the comments is the best part of it.

Splitting only on `", "` **because that is what every parser joins with** is the right call, and explicitly refusing a regex on `&` / `feat.` / `x` is what stops "Simon & Garfunkel" being shredded — a trap this exact feature falls into in most apps. Documenting the "Tyler, The Creator" limitation and using it to justify off-by-default is honest rather than hand-waved, and the tests cover the cases that actually matter, ampersand band names included.

Two notes. One is trivial; the second is a genuine gap and isn't your fault.

**1. The preference is re-read once per scrobble.** `scrobbleArtist` calls `sessionPreference.firstArtistOnly.firstOrNull()` on every invocation, and `drainQueue` loops `submit(...)` over up to 100 pending events — so a backlog drain does 100 DataStore reads of a value that cannot change mid-drain. Hoisting it to a single read in `drainQueue` and passing it down would be a one-line change. Minor, but backlogs after a cold-start import can be a few hundred.

**2. Loves are not covered, and they should be.** Last.fm love-mirroring landed this week (`LikeDestinationDispatcher`), and it calls `setLoved(artist = track.artist, ...)` with the **raw** artist string — lines 83 and 147.

So with your toggle on, a user gets:

- scrobble → `Calvin Harris`
- love → `Calvin Harris, Dua Lipa`

Two different artists on the same service for the same track. Last.fm matches loves against its own catalog entry, so the love may well land on nothing, or on a different entry than the scrobble — which is precisely the mismatch this toggle exists to prevent.

The fix is small: `primaryArtist` is already `internal` in `:core:data`, and `LikeDestinationDispatcher` is in the same module, so the dispatcher can apply the same transform behind the same preference. I'm happy to do that as a follow-up on my side since the love-mirroring is mine and landed after you opened this — just flagging it so we don't ship a half-consistent Last.fm identity.

Neither blocks merge. The core of this is right.

PR #395, opened 2026-07-31
Review pass 1: stop shipping developer machinery to users

rawnaldclark:
First pass of the codebase review. Class: **internal implementation leaking into the user-facing UI.**

This class was picked to go first because it is the only one with a *confirmed, already-paid* cost: a force toggle left enabled in a release install silently disabled lossless playback, and finding it took a debugging session that went as far as dex-dumping the shipped APK.

## Fixed

| Surface | Was | Now |
|---|---|---|
| `"Stream via amz (test)"` | Routed all streaming + downloads through a **parked** source | Deleted |
| `"Force Direct Qobuz only (test)"` | Shipped to every user | Debug builds only |
| Routing panel | `"amz.squid.wtf (Amazon Music) fills in when a track isn't on Qobuz"` | Names Qobuz; explains the YouTube fallback honestly |
| `StreamSourceRegistry` force-amz branch | Live | Parked with the source |

## The part that isn't cosmetic

Removing the amz toggle **alone would have stranded users**. Anyone who had switched it on still has `force_amz_only = true` in DataStore, and with the UI gone there is no way to switch it off — every track routed through a parked source, permanently, silently. So the registry branch had to go in the same change.

That is the general rule this pass establishes: **parking a source parks its force branch too.** It's the same failure shape as the force-YouTube incident — a stale preference quietly disabling lossless.

## Testing

1996 tests green. Two tests asserted the old amz-only routing; they're replaced by a guard for the actual hazard — *a stale `force_amz_only` preference is ignored and playback still resolves*. It deliberately stubs amz to **succeed**, so if the parked branch is ever restored the test fails instead of passing vacuously.

## Deliberately not touched

**ARCOD.** It has two surfaces: source plumbing (parked) and *partner attribution* on Home — Ko-fi link, Discord invite, the tagline "part of Stash's backbone". Removing a partner's attribution is a business decision, not a code-cleanliness one.

### DIFF
--- a/core/media/src/main/kotlin/com/stash/core/media/streaming/StreamSourceRegistry.kt
+++ b/core/media/src/main/kotlin/com/stash/core/media/streaming/StreamSourceRegistry.kt
@@ -112,11 +112,19 @@ class StreamSourceRegistry @Inject constructor(
                 // playlist would spend a search call + the user's arcod account
                 // on every queue track speculatively, not just the ones played.
                 if (allowYtDlp) add("arcod" to arcod::resolve)
-            } else if (streamingPreference.isForceAmzOnly()) {
-                // Test toggle (outage drill): amz ONLY — kennyy/squid/youtube
-                // removed from play so a track either streams via amz or fails
-                // visibly. Ignores allowYouTube/allowYtDlp — it's amz or nothing.
-                add("amz" to amz::resolve)
+                // NOTE: the force-amz branch is deliberately gone (2026-07-31).
+                //
+                // amz is parked, and its Settings toggle was removed with it — but a
+                // user who had already switched it on still has `force_amz_only = true`
+                // sitting in DataStore, with no UI left to switch it off. Had this
+                // branch survived, those users would route every track through a
+                // parked source and get silence, permanently, with no way out.
+                //
+                // Parking a source therefore has to mean parking its force branch in
+                // the same change. This is the exact failure shape that took a full
+                // debugging session when force-YouTube was left enabled in a release
+                // install: a stale preference quietly disabling lossless. The pref key
+                // and StreamingPreference accessor stay for re-enablement.
             } else if (streamingPreference.isForceYouTubeFallback()) {
                 // Test toggle: skip the lossless sources, forcing the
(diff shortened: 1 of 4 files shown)


rawnaldclark (commented):
Review record for pass 1 — scope, findings, and what was examined and left alone.

**Detector used**

```
grep -rhoE '"[^"]*\((test|beta)\)[^"]*"|"[^"]*(captcha|cookie|token|app_id|yt-dlp|InnerTube)[^"]*"'
grep -rhoE '"[^"]*(amz|qbdlx|Qobuz|arcod|squid|kennyy)[^"]*"'   # on feature/*/src/main
```

Anything a user can read that names our plumbing is a hit. Cheap to re-run each release; worth adding to CI later.

**Verified clean**

- The `"via YT"` badge in Now Playing — correct as-is. It names a service users recognise and explains why a track sounds different. That is the right kind of transparency; the goal is removing *machinery*, not honesty.
- `"Stream via YouTube"` — kept deliberately. Despite being the toggle behind the outage, it is a legitimate user escape hatch with honest help text, and its effect is already visible in Now Playing. The fix for that incident is that a *stale* pref can't strand you, not removing the control.

**Findings NOT fixed here** (deliberate — each needs a decision or a bigger change)

1. `"captcha_verified_at value"` as a text-field label, plus instructions to open devtools and copy a `Cookie` header. This is squid-captcha setup, and squid is parked — the whole surface may now be dead. Needs a check before deleting: it may already be unreachable UI.
2. `sp_dc` / `SAPISID` / `LOGIN_INFO` cookie flows for Spotify and YouTube auth. These leak machinery badly, but there is no alternative mechanism available — the honest fix is better presentation, not removal. Too large for this pass; deserves its own.
3. ARCOD source-connect UI (`"Connect ARCOD"`, `"ARCOD — connected"`, `ArcodConnectScreen`) for a parked source. Left alone because ARCOD is a partner, and that call is yours.

**Follow-up worth tracking:** the stranded-preference rule found here is general. Any pref that gates behaviour and can be removed from the UI has the same hazard — pass 4 (test integrity) should look for others.

PR #396, opened 2026-07-31
Review pass 2: stop dropping load-bearing events

rawnaldclark:
Second review pass. Class: **lossy async primitives carrying load-bearing events.**

## Signal-to-noise

The detector found **11 sites. 10 are correct.** Eight are `_userMessages` snackbars, one is a one-shot UI trigger, one is a summary display — all cases where `DROP_OLDEST` is exactly right, because the newest value supersedes the last. Reporting them would have been noise dressed as thoroughness.

One real defect.

## `_trackDeletions` could silently lose deletions

Emitted via `tryEmit` from five sites, all ignoring the returned `Boolean`. `tryEmit` never suspends — on a full buffer it returns `false` and the event is simply gone.

The consumer is `PlayerRepositoryImpl`, which evicts deleted tracks from the live queue. A dropped event therefore leaves a **deleted track playable, pointing at a file that no longer exists.** That symptom is not hypothetical — the comment at the collector cites a user report of exactly it.

`cleanOrphanedMixTracks()` and multi-select removal are precisely the bulk paths that can exceed the buffer.

**Fix:** all five call sites are `suspend` functions, so this becomes `emit` rather than a larger buffer. A slow consumer now makes the deleter wait instead of losing the notification. The buffer remains as headroom so an ordinary single delete never blocks.

## Also: correcting my own previous fix

`TrackIdentityEvents` got `extraBufferCapacity = 512` plus a warning log in the last release. That only makes loss *rarer* — it was still `tryEmit`, and I was guessing at a buffer size.

Its four call sites are all inside coroutines (`ensureYoutubeId`, `performSwap`, `canonicalize`, and the ViewModel's `viewModelScope.launch`), so `emitIdentityChanged` is now `suspend` and uses `emit`. Sized buffers make the failure rarer; suspending removes it.

Both flows carry cache/queue invalidation — every event load-bearing, none superseded by a later one. The exact inverse of the snackbar flows they were sitting next to, which is why the same primitive was right in ten places and wrong in two.

## Testing

1996 tests green across all modules.

### DIFF
--- a/core/data/src/main/kotlin/com/stash/core/data/repository/MusicRepositoryImpl.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/repository/MusicRepositoryImpl.kt
@@ -59,6 +59,22 @@ class MusicRepositoryImpl @Inject constructor(
     //
     // Buffer is generous so emits from a cascade-delete loop don't suspend
     // the caller (we use tryEmit).
+    /**
+     * Deleted track ids, consumed by PlayerRepositoryImpl to evict them from the
+     * live queue. Every event is load-bearing: a missed one leaves a deleted track
+     * playable, pointing at a file that no longer exists — the exact symptom a user
+     * reported (a removed song carrying on playing).
+     *
+     * Emitted with `emit`, NOT `tryEmit`. All five call sites are suspend functions,
+     * so real backpressure is available: a slow consumer makes the deleter wait
+     * instead of silently dropping the notification. `tryEmit` never suspends — on a
+     * full buffer it returns false, and all five sites ignored the result, so a bulk
+     * delete (`cleanOrphanedMixTracks`, a multi-select removal) could exceed the
+     * buffer and lose evictions with nothing logged.
+     *
+     * The buffer still exists so the common single-delete case never blocks; it is
+     * headroom now rather than the only thing standing between us and data loss.
+     */
     private val _trackDeletions = MutableSharedFlow<Long>(
         replay = 0,
         extraBufferCapacity = 64,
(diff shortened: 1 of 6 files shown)


rawnaldclark (commented):
Review record for pass 2.

**Detector**

```
grep -rn "BufferOverflow.DROP\|onBufferOverflow"        # dropping policies
grep -rn "tryEmit(" | grep -v "if (!\|val \|return "    # emits with ignored result
```

Then classify each hit by *what it carries*, which is the whole job. The primitive isn't wrong or right in itself — `DROP_OLDEST` is correct for a snackbar and catastrophic for a cache invalidation. 11 hits, 10 correct, 1 defect.

**Verified correct, left alone (10)**

`_userMessages` × 8 (PlayerRepositoryImpl, HomeViewModel, LibraryViewModel, AlbumDetailViewModel, ArtistDetailViewModel, LikedSongsDetailViewModel, PlaylistDetailViewModel, FailedMatchesViewModel), `_showStreamingDisclosure`, `_lastCascadeSummary`. All UI signals where the newest value supersedes; dropping an older snackbar is the desired behaviour.

**The rule this pass produces**

> Choose the overflow policy from the *payload*, not the call site. If a later event makes an earlier one redundant → `DROP_OLDEST`. If every event is load-bearing → suspend, and if you're reaching for a buffer size, you're guessing.

**Worth noting about the method**

The `trackDeletions` defect was found by a detector derived from a bug fixed the day before — `TrackIdentityEvents`, which literally documents itself as *"Mirrors MusicRepository.trackDeletions, same reasoning."* The comment was accurate: it inherited the reasoning **and** the defect. Fixing one instance of a class and not sweeping for siblings is how these persist.

**Follow-up not done here**

`Channel(...)` capacities weren't audited — `LikeCoordinator` uses `Channel.UNLIMITED` (safe), but I didn't sweep every `Channel(` construction for bounded ones carrying load-bearing work. Small, worth folding into a later pass.

Issue #368, opened 2026-07-24
[Bug]:


DONIFOL:
### Before you start

- [x] I'm on the **latest version** from the [Releases page](https://github.com/rawnaldclark/Stash/releases/latest)
- [x] Did you check the [task board](https://github.com/users/ParaliyzedEvo/projects/3/) for any **existing issues** that match your bug? (If you find one, comment on it instead of filing a new report.)
- [x] This is a **bug**, not a question or feature request (those go to [Discord](https://discord.gg/vcbjEby5PC))

### Stash version

V0.9.81

### Phone & Android version

Realms 15, Android  15

### Where does it happen?

Downloads (Offline mode)

### What happened, and what did you expect instead?

Syncs everything, even if I toggle off it in "manage playlist", I have like 500+ tracks. But it downloads 6000+ from users playlists and Spotify mixes that I don't need

### Steps to reproduce

<img width="1080" height="2392" alt="Image" src="https://github.com/user-attachments/assets/dbf030b0-f7eb-422e-a6d3-b01b8774d4c9" />

<img width="1080" height="2392" alt="Image" src="https://github.com/user-attachments/assets/301217b7-b985-40ad-869c-13bf2828ab13" />

PR #397, opened 2026-08-01
Review pass 3: orphaned guards — one dead feature removed, one trap documented

rawnaldclark:
Third review pass. Class: **orphaned guards — logic that exists and is never consulted.**

This class earned a dedicated sweep because it's the one that shipped #368: `shouldEnqueueForDownload` was written, unit-tested, and never wired, and users got thousands of downloads they never asked for.

## Detector 1 — unused `internal` functions

9 hits, **all `*ForTest` seams.** Intentional by naming convention; every one a false positive.

Worth recording rather than hiding: this detector *would* have caught #368 (`shouldEnqueueForDownload` was exactly this shape — `internal fun`, zero `src/main` references). That shape is now absent from the codebase.

## Detector 2 — preferences with no production consumer

6 hits, **three different verdicts.** Reporting them as one finding would have been wrong three times.

**Deleted — `LosslessInboxPreferences` (88 lines).** A DataStore for a watch-folder "lossless inbox". Zero references outside its own file; it was the only file in its package. Not a guard that stopped being called — a feature that never arrived.

**Documented, deliberately not deleted — `heartDefault*`.** These were the original "which services does a heart reach" controls. That job now belongs to `mirrorLikesSpotify` / `mirrorLikesYtMusic`, which `LikeCoordinator` actually reads. The heartDefault flows are still plumbed into `SettingsUiState`, but no switch renders them and nothing outside tests consults them.

They stay because **`SettingsViewModel` assembles UiState with an index-addressed combine** (`values[27]`, `values[28]`, …). Removing a source shifts every later index and hands the UI a type-correct wrong value — a worse bug than the dead weight, and one no test would catch. The comment records the actual trap: adding a Settings switch for these would give users a control that does nothing.

**Noted, untouched — `streamQuality`** in `StreamingPreference`. Also consumer-free, same index-combine reason. Flagged rather than half-removed.

## The finding behind the findings

Two of six hits are dead because of that index-addressed `combine`. It makes removing *any* UiState source unsafe, so dead sources accumulate — the structure actively resists cleanup. Converting it to named fields would be a real improvement and is the kind of thing a later pass should take on deliberately, not as a drive-by.

## Testing

1996 tests green. Net −70 lines.


rawnaldclark (commented):
Review record for pass 3.

**Detectors**

```
# 1. internal functions never referenced from src/main
#    (would have caught shouldEnqueueForDownload / #368)
# 2. preference accessors (val x: Flow<..>, suspend fun xNow()) with
#    zero references outside their declaring Preferences file
```

Both are cheap to re-run and neither needs the codebase loaded into context — worth CI-ifying later.

**Outcome: 15 hits, 1 deletion, 2 documented, 12 false positives.**

That ratio is the honest result and I'd rather state it than pad the pass. The class is largely clean; #368 was the outlier, not the pattern.

**What I did NOT do, and why**

Deleting `heartDefault*` and `streamQuality` is a two-line change that I chose not to make. `SettingsViewModel.uiState` is built from an index-addressed `combine` — `values[27] as Boolean`, `values[28] as Boolean`, and so on. Removing any source silently shifts every later index. The cast still succeeds, the type still checks, the UI just gets the wrong value. No test catches it because every field still has *a* value of the right type.

I hit this exact hazard adding ListenBrainz last week and worked around it by exposing standalone StateFlows instead of touching the combine. That workaround is now the third time this structure has bent someone's change around itself.

**Structural finding worth its own issue**

The index-addressed combine is *why* two of these six dead prefs still exist: the safe move is always to leave them. A structure that makes cleanup unsafe accumulates cruft by construction. Converting it to a named-field builder is a contained, mechanical refactor with a real payoff — and it should be its own change with its own verification, not a rider on a cleanup pass.

**Method note**

Pass 1 → 4 findings, pass 2 → 1 finding from 11 hits, pass 3 → 1 from 15. The yield is dropping, which is the expected shape: the failure classes were ranked by observed frequency, so the early passes should be the rich ones. If pass 4 (test integrity) also comes back thin, that is a real signal the targeted approach has been exhausted and the remaining value is in Pass 0 — the unstructured deep read that finds classes we haven't named yet.

PR #398, opened 2026-08-01
Remove the positional indices from the Settings uiState combine

rawnaldclark:
The structural finding from review pass 3. **Not a bug fix — removing a structure that manufactures bugs.**

## The problem

`SettingsUiState` is built from **38 flows**. Kotlin's `combine` is typed only up to five, so beyond that the lambda gets `Array<Any?>` and every value was read positionally:

```kotlin
val mirrorLikesSpotify = values[27] as Boolean
val mirrorLikesYtMusic  = values[28] as Boolean
```

The flow list and the read list were two parallel lists kept in sync **by hand**. Insert a source in the middle and every later index shifts — the cast still succeeds, the type still checks, and the UI quietly renders the wrong value. **No test catches it**, because every field still holds a valid value of the right type.

## Why it was worth doing

This isn't theoretical. It had already bent three separate changes around itself:

- `heartDefault*` survives as a dead preference because deleting it is unsafe
- `streamQuality` likewise
- ListenBrainz was exposed as standalone `StateFlow`s last week *specifically* to avoid touching this

A structure that makes cleanup unsafe accumulates cruft by construction. Two of pass 3's findings existed *because of* this.

## The change

All 38 positional reads become sequential `v.next<T>()` calls in declaration order. Adding a flow now means adding one call in the matching position — nothing to renumber. `requireExhausted()` turns the remaining mistake (a flow added or removed without a matching read) into an immediate, named failure instead of silently shifted data.

## What this does *not* fix

Reordering two same-typed flows is still silent. I'd rather say that than overclaim.

The real remedy is structural: this is a **hub-and-spoke** Settings UI, so each spoke should observe its own small typed state rather than every screen sharing one 38-source object — which would also stop every screen recomposing on unrelated changes. That's a much larger change, documented on the `Values` class rather than half-attempted here.

## Verification

Mechanical, order-preserving transformation. 1996 tests green.

**Not yet confirmed visually on device** — the phone was in use when I got here. Given this touches every value on every Settings screen, it's worth a look at Audio & Quality and Accounts before merging. I'd rather flag that than imply a check I didn't run.

### DIFF
--- a/feature/settings/src/main/kotlin/com/stash/feature/settings/SettingsViewModel.kt
+++ b/feature/settings/src/main/kotlin/com/stash/feature/settings/SettingsViewModel.kt
@@ -120,9 +120,10 @@ class SettingsViewModel @Inject constructor(
     // ── ListenBrainz ────────────────────────────────────────────────────────
     //
     // Exposed as standalone StateFlows rather than threaded through the big
-    // combine(...) below, which addresses its sources by numeric index
-    // (values[27], values[28]…). Adding four more inputs there would shift every
-    // later index — a silent, type-correct way to hand the UI the wrong value.
+    // combine(...) below. That combine no longer uses numeric indices (see
+    // [Values]), but it is still 38 sources feeding one monolithic UiState, and
+    // ListenBrainz belongs to one screen — keeping it separate is the right shape
+    // regardless.
 
     private val _listenBrainzTokenInput = MutableStateFlow("")
     val listenBrainzTokenInput: StateFlow<String> = _listenBrainzTokenInput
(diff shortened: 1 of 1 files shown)


rawnaldclark (commented):
Review record.

**Why a sequential reader rather than typed bundles**

The textbook fix is nested `combine`s with intermediate typed data classes — 38 sources into ~8 bundles of ≤5, then combining the bundles. Fully type-checked, and I started down that path before rejecting it.

It costs roughly 200 lines of bundle boilerplate to defend against index drift, on a class that is already 1503 lines. That trade is bad on its own terms, and worse given the boilerplate would itself need maintaining. The sequential reader removes the same failure mode (renumbering on insert/remove — the one that has actually bitten three times) in ~15 lines, and `requireExhausted()` makes the residual case loud instead of silent.

The honest limit is reordering two same-typed flows, which stays silent. I have documented it on the class rather than pretending otherwise, because the genuine fix for that is not more machinery here — it is per-spoke state.

**Why the monolith is the real issue**

Settings is hub-and-spoke: Playback, Audio & Quality, Accounts, Library & Storage, Appearance. Each spoke needs a handful of values. All five currently observe one 38-source `StateFlow`, so every screen recomposes whenever any unrelated preference changes, and any change to the shared object is a change to all of them.

Splitting per spoke would eliminate the vararg `combine` entirely — no bundles, no reader, no indices, each screen with ≤5 typed sources — and take the recomposition win with it. That is the change I would recommend next in this area, and it wants its own PR and its own verification rather than riding along here.

**Verification gap, stated plainly**

Compile + 1996 tests + a mechanical order-preserving transformation + `requireExhausted()` proving the count matches. What I do **not** have is eyes on the rendered screens; the phone was dozing and in use. Given the blast radius is "every value on every Settings screen", that check should happen before merge. I can run it as soon as the device is free.

PR #399, opened 2026-08-01
ARCOD: client for the /v2/stash API (not yet wired)

rawnaldclark:
Client for the API ARCOD's dev supplied. **Deliberately not wired into either registry** — see the last section.

## ARCOD probably wasn't a dead source. It was a dead client.

Our existing `ArcodClient` targets an API that no longer exists in that form:

| | Ours (parked) | The actual API |
|---|---|---|
| Host | `arcod.xyz/api` | `api.arcod.xyz` |
| Search | `/api/get-music?q=` | `/v2/stash/search?q=&limit=&offset=` |
| Stream | `POST /v2/downloads` → **poll** `/v2/downloads/<id>` until `completed` | `GET /v2/stash/stream/:id?quality=27` — one call |
| `X-Stash-Key` | **never sent** → automatic 403 | required on every route |

That mismatch is a far more likely explanation for ARCOD "not working" than the source being broken.

The current API is also **simpler than what we built against**: one GET returns a direct, Range-seekable FLAC URL. No job queue, no polling, no client-side decrypt — the same profile as qbdlx, which is what makes it the cheapest of the three lossless-redundancy options currently open (qbdlx is one live token deep; Tidal is written but unmerged and its proxy pool is unverified).

## Credential handling

`X-Stash-Key` is private. It follows the rule already established for `arcod.streamBase`: `local.properties` or a CI secret → `BuildConfig` → never committed. Verified before pushing that the key appears in **no tracked file and no commit on any branch**.

## Why this isn't wired in

Every `/v2/stash/…` call currently returns `403 {"error":"Forbidden"}` from ARCOD's own middleware — the body shape is hand-written, not Cloudflare's HTML block page. It's identical with and without the key, and never reaches the **401** the spec defines for a bad user token, so the request is refused *before* auth is evaluated. Either the key rotated (the spec is four weeks old) or the routes are IP-allowlisted. The operator has been asked.

Wiring a source chain onto an API that has never returned 200 is exactly the mistake made with Discord this week — a module scaffolded on the assumption a scope would be granted, which it wasn't. The client is pure spec translation and fully testable without the server, so that part is done; the wiring waits for a 200.

## Testing

10 contract tests, green. They lock in the two things we can be wrong about independently of the server:

- **What we send** — both required headers, the stash routes rather than the generic `/v2/search` ones the spec explicitly warns off, optional `Token-Country`
- **How we read documented responses** — particularly 401 vs 403, which look identical to a user and mean opposite things to us: reconnect vs the integration itself being refused

When the 403 clears, a green suite here means any remaining failure is the server's shape differing from its spec — a much smaller search space.

### DIFF
--- a/data/download/build.gradle.kts
+++ b/data/download/build.gradle.kts
@@ -28,6 +28,14 @@ val arcodLocalProperties = Properties().apply {
 val arcodStreamBase: String =
     arcodLocalProperties.getProperty("arcod.streamBase") ?: System.getenv("ARCOD_STREAM_BASE").orEmpty()
 
+// Private per-integration key the arcod operator issues for Stash's /v2/stash/*
+// routes, sent as the `X-Stash-Key` header. Shared on the condition it stays
+// private, so it lives in local.properties / a CI secret and is NEVER committed —
+// same rule as arcod.streamBase above. Empty is valid: ARCOD then reports itself
+// unconfigured and the source chain skips it.
+val arcodStashKey: String =
+    arcodLocalProperties.getProperty("arcod.stashKey") ?: System.getenv("ARCOD_STASH_KEY").orEmpty()
+
 // ── qbdlx (direct-Qobuz) credentials + token pool ──────────────────────────
 // Bundled at build time from local.properties / env. APP_ID + APP_SECRET are
 // public (shown on qbdlx's login page). TOKEN_POOL is a comma-separated list of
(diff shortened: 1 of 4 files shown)


rawnaldclark:
Superseded — the `/v2/stash` integration shipped in v0.9.90 via `feat/arcod-stash-relaunch`, which went further than this branch: the client is wired into both registries, `ARCOD_CONFIGURED` now keys off the private stash key, and the release workflow fails the build if that key is missing rather than shipping arcod silently disabled.

Verified on-device before release: a 23-hour-expired access token self-refreshed against Supabase unprompted, then `arcod served … quality=7`.

Closing rather than merging because this branch's base predates the qbdlx multi-credential pool work and merging it now would revert it.

Issue #380, opened 2026-07-30
[Bug]: Crash when trying to delete large playlist


AverageMintUser:
### Before you start

- [x] I'm on the **latest version** from the [Releases page](https://github.com/rawnaldclark/Stash/releases/latest)
- [x] Did you check the [task board](https://github.com/users/ParaliyzedEvo/projects/3/) for any **existing issues** that match your bug? (If you find one, comment on it instead of filing a new report.)
- [x] This is a **bug**, not a question or feature request (those go to [Discord](https://discord.gg/vcbjEby5PC))

### Stash version

v0.9.87

### Phone & Android version

moto g stylus 2023 5g, android 14

### Where does it happen?

Other / not sure

### What happened, and what did you expect instead?

[crash-2026-07-30_03-17-43.txt](https://github.com/user-attachments/files/30529879/crash-2026-07-30_03-17-43.txt)
[stash-diagnostics-1785381605994.txt](https://github.com/user-attachments/files/30529878/stash-diagnostics-1785381605994.txt)

### Steps to reproduce

1. have large playing (mines 6000 some songs)
2. try to delete it.
3. crash

ParaliyzedEvo:
@AverageMintUser Fixed in latest Stash version

PR #413, opened 2026-08-09
Sync improvements: library verification (reconciliation/disk-check/adoption), FLAC upgrade sweep, and split Playback/Download modes

### LINKED ISSUE #44: Option to convert to FLAC
Currently, due to many source and FLAC source downtime, library is having mixed file types. Can we have a button to download all low quality songs again using FLAC source?
Single click to convert.

---
_Imported from Discord: https://discord.com/channels/1508402965042561054/1522254639297335397_

### LINKED ISSUE #85: Seperate Playback Mode And Download Mode (Offline and Online)
Currently, setting offline and online acts on both downloading and streaming. So currently if i want to download some songs at the same time play my playlist which has both online and offline songs, i cant; because in order to play the online songs in the playlist i have to switch to the online mode.

I propose for there to be 2 settings, one for playback mode and download mode, and for each you can select either online or offline. With this you can have for example, online playback, but offline downloads or online playback but online downloads (streaming only).
Offline playback but online download would be playing the already downloaded songs (if any) but downloading would just create an index for streaming instead of downloading offline files.
Offline playback and offline download is for purely playing your already downloaded songs.
I think that cleared out some doubts. Thanks.

---
_Imported from Discord: https://discord.com/channels/1508402965042561054/1528324286031597678_

### LINKED ISSUE #105: Refetch tracks if the audio file somehow gets deleted
Check every song plus the path of the audio file when syncing, if it can't find the file from its given path, redownload it.

---
_Imported from Discord: https://discord.com/channels/1508402965042561054/1531141914609455244_

### LINKED ISSUE #361: [Feature]:  Separate Sync bars / progress screens for remote syncing of tracks vs indexing and reconciliation of local library
### Before you start

- [x] I [searched existing issues](https://github.com/rawnaldclark/Stash/issues?q=is%3Aissue) and nobody's already requested this

First of all thank you for all the amazing work you are doing, its nothing less than great!!

One additional UX enhancement that might help manage user expectations during large syncs (as I myself experienced this syncing 4k+ songs

### What problem does this solve?

Currently, the progress shown on the Sync screen appears to represent the entire synchronization process. However, a significant portion of that time can actually be spent scanning and reconciling the existing local library (checking whether tracks are already downloaded, verifying quality/lossless availability, etc.) before any new downloads begin.

This leads to an average of 40+ minutes of wait time before a single track even dowloads:

Image

### What would you like Stash to do?

It may be worth separating these into two distinct operations:

Sync Progress – dedicated to fetching metadata and downloading tracks that actually need to be synchronized.

Library Verification / Scan – a separate operation responsible for reconciling the local library, identifying already downloaded tracks, validating lossless availability (if applicable), and updating the local database.

From a UI perspective, the Library & Storage → Library section could expose this as a dedicated action with its own progress indicator, instead of combining it with the sync workflow.

This would make it much clearer to users whether the app is actively downloading new content or simply processing the existing library, reducing the perception that the sync has stalled when it is actually performing local reconciliation.

I have offered this suggestion because it doesn't just ask for another progress bar, it proposes separating two conceptually different tasks:

Remote synchronization (network/download work).
Local library reconciliation/indexing (filesystem/database work).

That distinction is common in certain other media applications (e.g., Plex, Jellyfin, Kodi, and music managers), and it makes the UI much easier to understand during long-running operations.

As an extension to this, would it make sense for the same library verification pass to also refresh the statistics shown on the Sync page (total tracks, Spotify/YouTube Music track counts, storage used, FLAC usage, etc.) by reconciling them against the actual files currently present on device storage? That would provide a dedicated mechanism for rebuilding those statistics if they ever become out of sync, without requiring a full synchronization.

### Anything else? (optional)

_No_Response_

ParaliyzedEvo:
## Summary
- `TrackDownloadWorker` inlined four housekeeping steps (orphan sweep, retry resets, requeue) at the start of every sync, with no visibility or standalone use. Extracted into `LibraryReconciliationUseCase` (`:core:data`), called by both the sync chain and a new standalone action.
- Added a disk-existence check to the same pass: tracks marked `is_downloaded = 1` are verified against actual storage, and any whose files are missing get reset so they're picked up by the requeue step. Needed a `FileOrganizer`-backed check, but `FileOrganizer` lives in `:data:download` (downstream of `:core:data`), so added `FileExistenceChecker`, a functional interface in `:core:data` bound to a real implementation via Hilt `@Module` in `:data:download` — same inversion pattern as `LibrarySizeHolder.refresh()`.
- Exposed the pass standalone via `LibraryHealthViewModel.runVerification()` and a new "Verify library" card (ViewModel-scoped, no WorkManager, mirroring `runBackfill()`).
- **Known limitations (unfixed at this point):** SAF disk checks could false-positive as "missing" because cached `content://` URIs don't reliably resolve after a tree re-grant; resulting resets weren't always successfully requeued. No data loss, but unreliable on SAF/external storage. Also out of scope: adopting on-disk files with no corresponding DB row.
- Added a FLAC-upgrade sweep, run unconditionally at the end of every sync (independent of REFRESH/ACCUMULATE), that re-attempts downloaded non-FLAC tracks through `TrackDownloader`. A file is only replaced if the result is `Success` **and** actually `flac` (lossy yt-dlp fallback successes are discarded). Logs `"$count song(s) upgraded to lossless"` to `SyncLog`.
- Split `StreamingPreference.enabled` (which conflated "sync downloads real files" with "Now Playing can stream undownloaded tracks") into two independent flags: `StreamingPreference` now governs Download Mode only; new `PlaybackModePreference` governs Playback Mode, read at all gating sites in `PlayerRepositoryImpl`. `SyncHeroCard` now shows two separate toggles; Sync button label ("Update Streaming Index" / "Download Tracks to Device") keys off Download Mode only.
- **Follow-up fixes from tester report:** a fresh install pointed at a folder of previously-downloaded files wasn't recognized, since Verify only reconciles rows already marked `is_downloaded = 1`. Root-caused alongside it: `FileOrganizer.fileExists()` relied on a stale cached SAF URI, causing false "missing" negatives after reinstall.
  - `FileOrganizer.checkExists()`/`resolveExistingSafFile()` now walk the *current* SAF tree grant using the same artist/album/title slug convention used to write files, healing the DB (`TrackDao.healFilePath`) when a file is found under a new URI.
  - New `AdoptExistingFilesUseCase` (`:data:download`): a second Verify pass that probes undownloaded, sync-enabled-playlist tracks for a matching on-disk file via the same slug-walk, and calls `markAsDownloaded` directly on a hit — no network. Deliberately skips the network-auth source gate used elsewhere, since this bug pattern was partly caused by that gate blocking requeue when a source wasn't connected.
  - Also noted (no code change): toggling a playlist's sync on doesn't retroactively backfill `playlist_tracks` — a second sync is needed before Verify/adoption has rows to work with.

## Changes
- **New:** `LibraryReconciliationUseCase.kt`, `FileExistenceChecker.kt` (`:core:data`); `FileExistenceCheckerModule.kt`, `AdoptExistingFilesUseCase.kt` (`:data:download`); `PlaybackModePreference.kt` (`:core:data`)
- `TrackDao.kt`: added `getDownloadedTrackRefs()`, `resetMissingFiles()`, `healFilePath()`, `getAdoptionCandidates()`, `getLosslessUpgradeCandidates()`, new projections `TrackExistenceRef`/`TrackAdoptionCandidate`
- `FileOrganizer.kt`: replaced simple `fileExists(path)` with `checkExists(artist, album, title, filePath): FileExistenceResult` + `resolveExistingSafFile(...)`, walking from the live tree grant instead of a cached URI
- `SyncPhase.kt`/`SyncStateManager.kt`: new `VerifyingLibrary` phase between Diffing and Downloading
- `TrackDownloadWorker.kt`: swaps inlined housekeeping for `reconciliationUseCase.reconcile(...)`; adds `runLosslessUpgradeSweep()` after the `PerfSummary` log line
- `PlayerRepositoryImpl.kt`: all playback-gating call sites switched from `streamingPreference` to `playbackModePreference` (`streamOnCellular` and downloaded-file-first checks unchanged)
- `SyncViewModel.kt`/`SyncHeroCard.kt`/`SyncScreen.kt`: new independent `playbackOnline` state/toggle alongside existing `downloadOnline`
- `LibraryHealthViewModel.kt`/`LibraryHealthScreen.kt`: `runVerification()` now runs reconciliation + adoption in one pass; `LibraryVerificationStatus.Done` reshaped to carry both results; UI updated accordingly
- Test updates: `SyncViewModelTest` gets `playbackModePreference` mock

Closes https://github.com/ParaliyzedEvo/Stash/issues/44 
Closes https://github.com/rawnaldclark/Stash/issues/361 
Closes https://github.com/ParaliyzedEvo/Stash/issues/85
Closes https://github.com/ParaliyzedEvo/Stash/issues/105

## Test plan
- [x] `./gradlew :feature:sync:kspDebugKotlin`
- [x] `./gradlew :core:data:kspDebugKotlin`
- [x] `./gradlew test` (`:core:data`, `:data:download`, `:feature:settings`)


rawnaldclark (changes requested):
There's a lot to like here — extracting the inlined housekeeping into `LibraryReconciliationUseCase` is overdue, the SAF slug-walk + one-shot `buildSafIndex()` design is genuinely good (the perf note about `findFile()`'s hidden linear scan is exactly right), and the honest "known limitations" section is appreciated. Suites run green locally on the branch. Three findings need resolving before merge, though — the first two are ship-blockers in my view:

**1. Streaming mode now drains the download queue (regression).** The old `doWork` had an early return after housekeeping when `streamingPreference.current()` — with the long v0.9.30 comment explaining that pre-toggle PENDING rows must *not* drain in Online mode. This PR deletes that return, and after the reconcile `if/else` the code falls straight through to "Collect ALL pending items" + the parallel download loop. Net effect: with Download = Online, tapping "Update Streaming Index" downloads real files whenever stale PENDING rows exist — contradicting the exact semantics this PR introduces. The early return (or an equivalent gate before the drain) needs to come back.

**2. The FLAC-upgrade sweep is an unbounded retry herd.** Two parts:
- `runLosslessUpgradeSweep` launches one coroutine **per candidate** inside `supervisorScope` with no semaphore — everything else in this worker (and your own `AdoptExistingFilesUseCase` in this same PR) bounds concurrency at 8. A user with a few thousand lossy tracks fires a few thousand concurrent `TrackDownloader` calls at the lossless sources in one shot. We've had a real thundering-herd incident before (the v0.9.62 amz mint herd), and right now the primary lossless source is a partner's box — this pattern can take it down.
- There's no attempt-stamping or backoff: the same candidates that failed to find a lossless source get re-attempted on *every sync, forever* (the kdoc even says "expected for most candidates most syncs"). Master already has the infrastructure this wants: `FlacUpgradeQueueDao` — the status-tracked worklist the batch FLAC-upgrade worker uses (spec 2026-07-22 §3). The sweep should enqueue through that (or explain in the PR why it can't), which gives bounded work, per-track status, and no permanent retry loop for free.

**3. The Playback/Download split strands existing Online-mode users.** `PlaybackModePreference` defaults to `false` with the comment "matching pre-existing behavior" — but for the cohort with `StreamingPreference.enabled = true`, pre-existing behavior was that Now Playing streams. After this update their playback silently stops streaming until they discover the new toggle. A one-time seed (initialize `playback_online_enabled` from `StreamingPreference.enabled` when the key is absent) preserves everyone's current behavior and costs a few lines.

**Smaller things:**
- `AdoptExistingFilesUseCase.kt` lives in `data/download/.../files/` but declares `package com.stash.core.data.library` — package and directory disagree.
- The reconcile requeue inserts `DownloadQueueEntity(trackId, syncId = null)` where the old inline code stamped the real `syncId` — sync-attribution in diagnostics/history loses those rows.
- `SyncStateManager.onVerifyingLibrary(total = 4)` default disagrees with `TOTAL_STEPS = 5` (harmless today since the call site passes it, but it'll bite the next caller).
- A few files are missing trailing newlines.

Also worth saying: the new TrackDao reads are one-shot `suspend` lists, so they don't hit the CursorWindow refill race we just fixed in #380 for the reactive feeds — no change needed, just noting it was checked.

Happy to re-review fast — 1 and 3 look like small diffs, and 2 mostly becomes a call into infrastructure that already exists.

ParaliyzedEvo:
@rawnaldclark this should be good to go now. let me know if here are any additional changes you want me to do. i did all the main fixes plus the smaller ones other than the inline ones

rawnaldclark (approved):
Re-reviewed the revision — all three blockers are properly resolved:

1. **Streaming-mode drain**: early return restored with the v0.9.30 rationale carried forward in the comment. One small behavior delta vs the pre-PR code, fine to leave as-is but noting it: the old flow still ran the orphan-sweep/stale-reset housekeeping in streaming mode before returning; now the whole reconcile is skipped, so online-mode users' queue counters only self-heal on an offline sync or a manual Verify. Acceptable given Verify now exists — just don't be surprised by stale counters in that mode.
2. **FLAC sweep through `FlacUpgradeQueueDao` + `FlacUpgradeWorker`**: exactly right, and the `cancelProgress` → `cancelFlacUpgrade` notification-ID fix you found along the way is a genuine catch. One non-blocking follow-up worth an issue: `getLosslessUpgradeCandidates()` still returns every lossy track each sync, so a big lossy library re-attempts its full list (paced, but still source traffic). A `last_upgrade_attempt_at` stamp + "skip if attempted < N days ago" filter would cut steady-state traffic to near zero.
3. **`seedFromLegacyIfAbsent`**: precisely the shape needed — key-presence gated, never overwrites.

Package fix, syncId threading, and the total=5 default all verified. Suites green locally on the head (`:core:data`, `:feature:settings`, `:feature:sync`, app compile). Merging — thanks for the fast, thorough turnaround.

PR #414, opened 2026-08-09
feat: add JioSaavn AAC fallback before YouTube

### LINKED ISSUE #104: one folder
My friend really likes stash but he is hesitant to use it because he needs all the tracks in a single folder, so it would be nice if there was an option to select whether they are saved in separate folders, that is, by artist, album or ft, all the tracks in a single folder, since it is difficult for him to move them when there are especially many tracks

ParaliyzedEvo:
## Summary
- add a conservative JioSaavn 320 kbps AAC fallback after lossless sources and before YouTube for downloads and foreground streaming
- validate codec, bitrate, and duration before committing m4a files while preserving YouTube fallback for pre-commit misses
- add cancellation-aware, rate-limited, no-redirect transport, source attribution, and regression coverage
- prevent shared search-download cancellation races and post-commit double downloads

## Verification
- data download JioSaavn tests and SearchDownloadCoordinatorCompletionTest
- media StreamSourceRegistryTest, LazyResolvingDataSourceTest, and StashMediaSourceFactoryTest
- app compileDebugKotlin
- git diff --check

## Merge blocker
This integration uses an unofficial JioSaavn endpoint and media URL derivation. The current JioSaavn Terms of Use prohibit automated or unofficial access unless separately allowed: https://corporate.saavn.com/terms/. Confirm written provider authorization and applicable licensing before marking this ready to merge.

## Note
The existing force-YouTube diagnostic toggle behavior for already-resolved timeline items remains unchanged and is outside this PR.

Closes #104

Original PR by @JoshDui

### DIFF
--- a/core/media/src/main/kotlin/com/stash/core/media/service/StashPlaybackService.kt
+++ b/core/media/src/main/kotlin/com/stash/core/media/service/StashPlaybackService.kt
@@ -404,6 +404,15 @@ class StashPlaybackService : MediaLibraryService() {
                 val origin = item.mediaMetadata.extras?.getString(EXTRA_STREAM_ORIGIN)
                 (scheme == "http" || scheme == "https") && origin == "amz"
             },
+            // JioSaavn media is accepted only from its exact HTTPS CDN host.
+            // Its playback client disables redirects so a later range request
+            // cannot pivot from that validated host to another destination.
+            isJioSaavnOrigin = { item ->
+                val scheme = item.localConfiguration?.uri?.scheme?.lowercase()
+                val origin = item.mediaMetadata.extras?.getString(EXTRA_STREAM_ORIGIN)
+                (scheme == "http" || scheme == "https") &&
+                    origin == com.stash.core.media.streaming.JioSaavnStreamResolver.ORIGIN
+            },
             amzHttpClient = okHttpClient,
             // Full-timeline queue: stash-resolve:// placeholders resolve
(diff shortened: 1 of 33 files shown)


rawnaldclark:
Maintainer update — I've reworked the branch (thanks for allowing edits):

- **Rebased onto master.** The Spotify owner-data commit was already merged as #415, and the `DatabaseBackupManager` fix is also on master — both dropped out, which clears the merge conflict.
- **Split out the "Improve sync" commit** (library reconciliation / adopt-existing-files / Verify UI). It's genuinely useful work but it's a second feature with its own review surface — and as written it reintroduced the v0.9.30 streaming-mode bug (the early return that keeps Online-mode syncs from draining pre-toggle PENDING rows was removed). I've preserved it locally and we'll follow up on it as its own PR with that fixed.
- The PR is now **pure JioSaavn** (~1,650 additions): client/matcher/resolver, download + search + streaming wiring, tests. All targeted unit tests pass and the app compiles; on-device verification (logcat `origin=jiosaavn` + the "via JioSaavn" badge) is the remaining gate before merge.

Review notes on the JioSaavn core itself: the matcher's conservatism is exactly right (JioSaavn returns cover-floods for unlicensed Western artists — verified live), and the no-redirect host-pinned transport + ftyp probe are solid. One thing to watch post-merge: the hard explicit-flag veto may cost real matches where JioSaavn's `explicit_content` field is unreliable; if we see misses, we'll demote it to a score penalty.

PR #418, opened 2026-08-09
Persist Library sort order and source filter across restarts

### LINKED ISSUE #88: Stash keeps reverting music filter back to FLAC only
Hey,

I'm having an issue with Stash. By default it only shows FLAC music. When I change the filter to Show All Music, it works temporarily, and when i close the app and reopen it, it automatically reverts back to FLAC Only.

This is getting really annoying. Is this a bug or am I missing a setting? Is there a way to make Show All Music the default so it doesn't keep resetting?

Thanks!

---
_Imported from Discord: https://discord.com/channels/1508402965042561054/1528545866242330766_

ParaliyzedEvo:
## Summary
- The Library screen's sort order and source filter (`SortOrder`, `SourceFilter`) previously lived only in `LibraryViewModel`'s in-memory `ControlState`, so both silently reset to their defaults (`RECENT` / `ALL`) every time the app process was killed and restarted.
- Added `LibraryPreferencesStore`, a small Preferences DataStore wrapper (mirrors `core/media`'s `PlaybackStateStore` shape) that persists both values as their enum `.name` strings.
- `LibraryViewModel` now seeds `_controls` from the store on init (fire-and-forget `viewModelScope.launch`, so there's a brief one-frame flash of defaults before the persisted values land — same tolerated pattern as `NowPlayingViewModel.ambientAnimationEnabled` starting null) and writes through to the store on every `setSortOrder` / `setSourceFilter` call.
- `feature:library` didn't previously depend on `androidx.datastore:datastore-preferences` (only `core:media` did); added the dependency so the new store compiles.

## Changes
- `feature/library/src/main/kotlin/com/stash/feature/library/LibraryPreferencesStore.kt` (new)
  - `getSortOrder()` / `getSourceFilter()` — one-shot reads, defaulting to `SortOrder.RECENT` / `SourceFilter.ALL` on missing or unparseable stored values
  - `setSortOrder()` / `setSourceFilter()` — write-through on user change
- `feature/library/src/main/kotlin/com/stash/feature/library/LibraryViewModel.kt`
  - Injected `LibraryPreferencesStore`
  - Added `init` block seeding `_controls` from the persisted sort order + source filter
  - `setSortOrder` / `setSourceFilter` now write through to `LibraryPreferencesStore` in addition to updating `_controls`
- `feature/library/build.gradle.kts`
  - Added `implementation(libs.androidx.datastore.preferences)` (matches the version already pinned in `core/media/build.gradle.kts`)

## Test plan
- [x] `./gradlew :feature:library:compileDebugKotlin`

Closes https://github.com/ParaliyzedEvo/Stash/issues/88

---

*Maintainer addendum (scope note for the changelog):* beyond the Library sort/filter persistence described above, this PR also (a) persists **repeat mode** in `PlaybackStateStore` and restores it in `resumeLastQueue`, and (b) introduces **`PlaybackSource`** ("Playing from …") in `core/model`, persisted with the queue and shown in the queue-sheet header. `PlaybackSource` coverage is Library-initiated queues in this increment; playlist-detail/home/mix taps still resolve as `Unknown` (label hidden) and can be wired up incrementally.


rawnaldclark (changes requested):
The features here are all wanted — sort/filter persistence closes a real annoyance, and repeat-mode + "Playing from" persistence make resume feel finished. But this needs another pass before it can land:

**1. The branch fails its own test suites (9 failures).** `./gradlew :feature:library:testDebugUnitTest` on this branch:

- `LibraryViewModelSortTest` ×2, `LibraryViewModelSourceFilterTest` ×1 — these fail because of a real design bug, not stale fixtures: the `init { viewModelScope.launch { … _controls.update … } }` seed is fire-and-forget, so it lands *after* a `setSortOrder()`/`setSourceFilter()` call and clobbers the user's change back to the persisted value. The test dispatcher makes the race deterministic, but it exists in production too (fast user, slow first DataStore read). Suggested shape: only apply the seed if `_controls` still holds the defaults, or make the store the backing source of truth (a flow the controls derive from) instead of a one-shot copy.
- `LikedSongsDetailViewModelTest` ×1, `MixOfflineTapGuardTest` ×5 — these stub/verify the old 2-arg `setQueue(tracks, index)` and weren't updated for the new `source` parameter (the other six test files were).

The PR's test plan lists `:feature:library:compileDebugKotlin` only — for a change that touches the persisted playback-state schema, please run the module test suites and paste results the way #419 did.

**2. The body describes one of three features.** The commits also add (a) repeat-mode persistence in `PlaybackStateStore` + restore in `resumeLastQueue`, and (b) a new `PlaybackSource` sealed class in `core/model`, threaded through `PlayerState`, the queue-save gate, and the queue-sheet header UI. Both look reasonable (the `toPlayerRepeatMode` inverse mapper and the drain-gate extension are correct as far as I can trace), but a reviewer deciding on a persisted-schema change needs the summary to say so. Please update the description — and note that `PlaybackSource` coverage is Library-initiated queues only for now (playlist-detail, home, and mix taps still land as `Unknown`), which is fine as an increment if it's stated.

**3. Nits:** `PlaybackSource.kt` / `PlayerState.kt` / `LibraryPreferencesStore.kt` are missing trailing newlines; the mock stub in `LibraryViewModelSortTest` has a mis-indented line; `Closes …/issues/88` points at the fork's issue tracker rather than this repo.

Happy to re-review quickly once the suites are green — the underlying work is good.

ParaliyzedEvo:
@rawnaldclark this should be good to go now. let me know if here are any additional changes you want me to do. also CI does pass as you can see in this [run](https://github.com/ParaliyzedEvo/Stash/actions/runs/31448059582), idk why it doesnt pass here

rawnaldclark (approved):
Re-reviewed — the per-field `sortOrderUserSet`/`sourceFilterUserSet` guard is the right resolution for the seed race (flags set synchronously in the setters, checked inside the seed's update; both paths run on Main so the ordering is sound and `@Volatile` is cheap insurance). The two missed test files are updated and the full `:feature:library` + `:core:media` suites now run green locally, app compiles.

I appended a short scope addendum to the PR body covering the repeat-mode and `PlaybackSource` halves so the changelog stays honest — your original text is untouched.

Merging — thanks for the quick turnaround.

Issue #42, opened 2026-05-11
Playlists not showing as playlists


4M3L1407:
Hello,

First off I want to say a massive thankyou this app is absolutly amazing!

One issue I have noticed is that when I download songs from the Stash app I can't add them to playlists imported from spotify meaning I have to go back to spotify, add them there, and then re-sync.

It would be nice to be able to add songs added in stash to my imported playlists.

Thanks in advance :)

rawnaldclark:
i agree! it will be added soon

rawnaldclark:
Fixed in v0.9.24 (commit e4ae6c3). Imported Spotify and YouTube Music playlists now appear in the "Save to Playlist" picker, and your additions persist across re-syncs (new `locally_added` flag on playlist_tracks preserves user-added rows when REFRESH mode would otherwise rebuild the playlist from the upstream source). APK will be available in Releases once the workflow finishes building.

PR #419, opened 2026-08-10
fix(spotify): order Liked Songs by Spotify's add date (#410)

### LINKED ISSUE #410: [Bug]: old liked songs moving up to the top of the playlist
### Before you start

### Stash version

v0.9.90

### Phone & Android version

Samsung A35, Android 16

### Where does it happen?

Library

When I sync the Spotify's liked songs playlist some of the old songs from the bottom get readded and moved to the top

### Steps to reproduce

1. Sync the playlists
2. Check the liked songs

Image

Euphoria by Kendrick Lamar was added like 2 years ago, yet it's at the top of the list

chaudhary-lakshay:
Closes #410 — old Liked Songs jumping to the top of the list after a sync.

## Root cause

Liked Songs order is the raw array index of one paginated GraphQL fetch, and nothing more.

1. `TrackDao.getByPlaylist` orders `LIKED_SONGS` purely on `pt.position ASC` — only `DAILY_MIX` gets a timestamp tiebreak.
2. `position` is assigned in `PlaylistFetchWorker.fetchSpotifyPlaylists` as `mapIndexedNotNull { index, … }` — the index in the concatenated page results.
3. `SpotifyApiClient.getLikedSongs` calls `fetchLibraryTracks` with variables `{offset, limit}` only — **no `order`**. (Compare `getUserPlaylists`, which does send `put("order", JsonNull)`.) So that index reflects whatever the server returned for that account at that moment.
4. `PlaylistDao.insertAllCrossRefs` is `OnConflictStrategy.REPLACE`, so every sync restamps every synced row's position.

Whenever the returned order differs from add-date-descending — server default shifting under an order-less query, offset pagination over a list that moved, a partial page — an old track lands at a low index and the REPLACE write hoists it to the top.

Spotify does report a per-item save date, and it was never requested, parsed or stored. `grep -rn "added_at" --include=*.kt` finds only the local `playlist_tracks.added_at` column, which is insert time.

## Fix

Parse `items[].addedAt.isoString` and sort newest-first **before** the index is assigned, so `position` carries the intended order.

No new column, no Room migration, no DAO change. Sorting ahead of the index reuses `position` as the carrier. A persisted `remote_added_at` column would only be needed if the order had to survive independently of fetch order, which nothing currently requires.

Degrades safely:

- Items with no parseable date sink to the end, keeping their relative order (stable sort).
- A page with **no** dates at all is returned by identity, so if the endpoint stops reporting them the behaviour falls back to today's rather than scrambling.
- `parseLibraryAddedAt` accepts nested `addedAt.isoString`, a flat `addedAt` string, and Web-API `added_at`, returning null rather than guessing.

### Why the local column can't be the sort key

In REFRESH mode `clearSyncedPlaylistTracks` runs at `DiffWorker.kt:445`, *before* `existingCrossRefs` loads at `:574`. Every sync-owned row is therefore restamped `Instant.now()` on a full sync. The sort key has to be the remote timestamp.

## Evidence for the response shape

The add date sits on the **item wrapper**, a sibling of `track` — not inside the track object. Confirmed against persisted-query hash `087278b2…f944240` (the value already in `SpotifyAuthConfig.HASH_FETCH_LIBRARY_TRACKS`) as used by independent clients:

```
data.me.library.tracks
├── totalCount, pagingInfo
└── items[]
    ├── addedAt.isoString : ISO-8601
    └── track._uri / track.data.{name, uri, trackDuration, artists, albumOfTrack}
```

- `fabiogaliano/hearted.music` — `extensions/SPOTIFY-API.md` documents this operation by name and by the same hash, with the structure above.
- `sceneq/myutils` — `spotify-favorites-extract.jq` extracts `.data.me.library.tracks.items[].addedAt.isoString` from a live capture.
- Corroborating `addedAt.isoString` on the same private API: `LuluHuman/Spotify-internal-js`, `futo-org/grayjay-plugin-spotify`, `satsfy/spotify-get`, `AliAkhtari78/SpotifyScraper`.

Every one of those clients sends `{offset, limit}` and none sends `order`, which independently confirms that trusting the returned index is unsafe by construction.

## Refactor included

`parseLibraryTracksResponse` moves from a `private` method on `SpotifyApiClient` to a top-level `internal` function in `SpotifyTrackParser.kt`.

It is a pure `JsonObject -> List<SpotifyTrackItem>` but was untestable on a class requiring OkHttp and auth to construct — meaning the JSON contract this fix depends on had no test at all. Its new home matches the existing convention of top-level parsers with their own test class (`parseLibraryPage` / `SpotifyLibraryParserTest`). Same package, so the call site is unchanged.

## Tests

New fixture `fetch_library_tracks_liked.json` uses the documented shape with the array **deliberately mis-ordered** (2019, 2026, 2023) — that mis-ordering is the bug.

- `SpotifyLibraryTracksShapeTest` (9 tests) — the documented path, `addedAt.isoString` read off the wrapper, the sort correcting the order, metadata parsed alongside the date, the `libraryTracks` alias path, non-track URIs skipped, unrecognised shapes returning empty rather than throwing.
- `SpotifyLikedSongsOrderTest` (6 tests) — the ordering function directly, including undated and unparseable values sinking to the end and the identity return when no dates are present.

Results:

| Task | Result |
|---|---|
| `:data:spotify:testDebugUnitTest` | 40/40, 0 failures |
| `:core:data:testDebugUnitTest` | 657/657, 0 failures (142 suites) |
| `:data:spotify:compileDebugKotlin` | BUILD SUCCESSFUL |

An early run was 5/6 and caught a real inversion: `compareByDescending` flips the comparator's arguments, so `nullsLast()` placed undated tracks *first*. Fixed to `nullsFirst()`, with a comment so it doesn't get "corrected" back.

## Not verified

Stated plainly rather than implied:

- **End-to-end on a real account.** Reproducing #410 needs an `sp_dc` session with a multi-year like history; I don't have one. The change is unit-green and compiles, and it installs and runs on an emulator, but the emulator has no linked account so the ordering itself was never exercised against live data.
- **The response shape is third-party evidence**, not a live capture from this app. Two independent sources document it for this exact hash and the parser now asserts it, but a logcat from a signed-in device would be stronger. If the live field name differs, the identity-return path means behaviour falls back to current rather than breaking.

## YouTube Music

No equivalent fix is possible there. `data/ytmusic/src/test/resources/fixtures/liked_songs_page1.json` — a real InnerTube capture already in the repo — carries no per-item add date at all (a `grep -i date` scores hits only on substrings of `entityBatchUpdate` / `frameworkUpdates` / `musicLibraryStatusUpdateCommand`). There is no sort key available.

This PR does not touch that path — `sortLikedSongsByAddedAtDesc` is called only from `fetchSpotifyPlaylists`. If #410 is ever reported on YouTube, the fix would have to be "stop restamping positions", i.e. preserve existing `position` for rows already known locally.

## Separate issues noticed, deliberately not fixed here

Flagging rather than scope-creeping:

- **`locallyAdded` is wiped on every sync.** `DiffWorker.addCrossRefIfNotSoftDeleted` (`:752-757`) rebuilds the cross-ref without carrying `storedPrior.locallyAdded`, so the REPLACE insert clears it. The next REFRESH's `clearSyncedPlaylistTracks` (`locally_added = 0`) then deletes the row as sync-owned — which defeats the #42 protection, and may be the literal "get readded" behaviour reporters describe. Fix would be `locallyAdded = storedPrior?.locallyAdded ?: false`.
- **Silent truncation.** `PlaylistFetchWorker.kt:521` — `SyncResult.Empty` mid-pagination `break`s without setting `likedFetchPartial`, and `parseLibraryTracksResponse` returns `emptyList()` on any unrecognised shape. A parse failure on page 3 snapshots pages 1-2 as the complete library while `snapshotUnreliable` reports clean. Same shape as #343, one layer up.


rawnaldclark (approved):
This is a model PR — root-cause chain, third-party evidence for a private API shape, honest "not verified" section, and a fixture that encodes the bug itself (the deliberately mis-ordered array). Thank you.

Verified locally on the branch: `:data:spotify:testDebugUnitTest` + `:core:data:compileDebugKotlin` green. I also traced the `compareByDescending(nullsFirst())` comparator by hand — the argument flip is subtle and your comment explaining why it's `nullsFirst` (not `nullsLast`) is exactly the kind of note that stops a future "fix". The identity-return fallback means the worst case if Spotify drops `addedAt` is today's behavior, not a scrambled library — good degradation design.

The two flagged-not-fixed adjacencies (`locallyAdded` wiped on every sync, silent truncation on mid-pagination parse failure) are real and worth their own issues — if you file them I'll label them, otherwise I will.

Merging. Live-account confirmation will come from the #410 reporter — the safe-degradation path makes that an acceptable order of operations.

PR #430, opened 2026-08-16
Gate FLAC upgrade sweep on streaming/lossless settings and fix redownloads orphaning files into a new album folder

### LINKED ISSUE #219: Upgrade to flac flaw
Deletes the non flac files when trying to download them. first find the flac file, then delete then change path

---
_Imported from Discord: https://discord.com/channels/1508402965042561054/1536984955039776768_

### LINKED ISSUE #230: Download issues
The app download everything in normal quality then it downloads it in flac despite of me wanting the noemal quality only.

---
_Imported from Discord: https://discord.com/channels/1508402965042561054/1538906149753982986_

ParaliyzedEvo:
## Summary

* `TrackDownloadWorker.runLosslessUpgradeSweep()` populated `flac_upgrade_queue` and kicked off `FlacUpgradeWorker` unconditionally at the end of every offline sync, with no check against the Settings **"Lossless downloads"** master toggle — so the upgrade queue could still be built and drained even when the user had lossless off entirely.
* Added `LosslessUpgrader.isLosslessEnabled()` (backed by `LosslessSourcePreferences.enabled`) and gated `runLosslessUpgradeSweep()` on it, alongside the existing streaming-mode check.
* The sweep also re-attempted an upgrade on tracks the same sync run had just finished downloading — wasted, redundant lossless-source calls immediately after those tracks already went through the primary download path. `TrackDownloadWorker` now tracks per-run downloaded track IDs and excludes them from the upgrade candidate set.
* Separately: `FileOrganizer.commitDownload()` derived a track's destination folder from its *current* `album` field at commit time. When a track's `album` was backfilled after its original download (e.g. blank → populated by `updateAlbumIfEmpty`/`updateCanonicalMetadata`), a later redownload of that same track computed a different folder (e.g. `singles/` → `that-way/`) and silently orphaned the original file rather than overwriting it — visible as duplicate folders on disk and, per user report, occasional downgrades from FLAC back to a lossy format on the "new" copy since the redownload isn't guaranteed to hit the lossless path again.
* Fixed by having internal-storage commits first scan the artist's existing subfolders for a file already matching the title's slug and reusing that location before falling back to deriving a fresh path from the current `album` field.

## Changes

* `core/data/src/main/kotlin/com/stash/core/data/lossless/LosslessUpgrader.kt`

  * Added `isLosslessEnabled(): Boolean` to the interface.
* `data/download/src/main/kotlin/com/stash/data/download/lossless/LosslessUpgraderImpl.kt`

  * Injected `LosslessSourcePreferences`.
  * Implemented `isLosslessEnabled()`.
* `core/data/src/main/kotlin/com/stash/core/data/sync/workers/TrackDownloadWorker.kt`

  * Injected `LosslessUpgrader`.
  * Track successfully downloaded track IDs for the run (`downloadedTrackIds`).
  * `runLosslessUpgradeSweep()` now checks `losslessUpgrader.isLosslessEnabled()` and filters `recentlyDownloadedTrackIds` out of the candidate set before calling `startBatch`.
* `data/download/src/main/kotlin/com/stash/data/download/files/FileOrganizer.kt`

  * `commitDownload()` internal-storage branch now calls new `resolveInternalTargetFile()` instead of `getTrackFile()` directly.
  * `resolveInternalTargetFile()` scans existing album subfolders under the artist directory for a file whose name (minus extension) matches the title slug, reusing it if found; falls back to the existing `getTrackFile()` derivation otherwise.
* Test fixture updates:

  * `FlacUpgradeWorkerTest.kt` — both inline `LosslessUpgrader` fakes updated to implement `isLosslessEnabled()`.
  * `LosslessUpgraderImplTest.kt` — added `losslessPrefs` mock and updated `subject` construction.

## Test plan

* [x] `./gradlew :core:data:compileDebugUnitTestKotlin`
* [x] `./gradlew :data:download:compileDebugUnitTestKotlin`
* [x] `./gradlew :core:data:testDebugUnitTest`
* [x] `./gradlew :data:download:testDebugUnitTest`
Closes https://github.com/ParaliyzedEvo/Stash/issues/219
Closes https://github.com/ParaliyzedEvo/Stash/issues/230

### DIFF
--- a/core/data/src/main/kotlin/com/stash/core/data/lossless/LosslessUpgrader.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/lossless/LosslessUpgrader.kt
@@ -18,5 +18,8 @@ import com.stash.core.model.UpgradeResult
  * their default settings.
  */
 interface LosslessUpgrader {
+    /** Whether automatic lossless downloads are enabled by the master preference. */
+    suspend fun isLosslessEnabled(): Boolean
+
     suspend fun upgradeToLossless(track: Track): UpgradeResult
 }
--- a/core/data/src/main/kotlin/com/stash/core/data/sync/workers/FlacUpgradeWorker.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/sync/workers/FlacUpgradeWorker.kt
@@ -47,6 +47,15 @@ class FlacUpgradeWorker @AssistedInject constructor(
         createForegroundInfo(text = "Preparing…", progress = -1f)
 
     override suspend fun doWork(): Result {
+        // The preference can change after TrackDownloadWorker builds the
+        // persisted batch but before WorkManager starts this worker. Enforce
+        // consent again at execution time and discard the unstarted worklist.
+        if (!losslessUpgrader.isLosslessEnabled()) {
+            queueDao.clearPending()
+            Log.i(TAG, "Lossless disabled in Settings: discarded pending FLAC upgrades")
+            return Result.success()
+        }
+
         val pending = queueDao.pendingTrackIds()
         if (pending.isEmpty()) return Result.success()
         val total = queueDao.countAll()
--- a/core/data/src/main/kotlin/com/stash/core/data/sync/workers/FlacUpgradeWorker.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/sync/workers/FlacUpgradeWorker.kt
@@ -49,10 +49,14 @@ class FlacUpgradeWorker @AssistedInject constructor(
     override suspend fun doWork(): Result {
         // The preference can change after TrackDownloadWorker builds the
         // persisted batch but before WorkManager starts this worker. Enforce
-        // consent again at execution time and discard the unstarted worklist.
-        if (!losslessUpgrader.isLosslessEnabled()) {
+        // consent again at execution time and discard the unstarted worklist —
+        // but ONLY for the automatic sweep. This worker has two producers, and
+        // the user's explicit "Upgrade to FLAC" selection (FlacUpgradeEnqueuer,
+        // untagged) is a manual override that must run with the master toggle
+        // off, same contract as Now Playing's forced single-track upgrade.
+        if (inputData.getBoolean(KEY_AUTO_SWEEP, false) && !losslessUpgrader.isLosslessEnabled()) {
             queueDao.clearPending()
-            Log.i(TAG, "Lossless disabled in Settings: discarded pending FLAC upgrades")
+            Log.i(TAG, "Lossless disabled in Settings: discarded pending auto-sweep FLAC upgrades")
             return Result.success()
         }
 
@@ -130,6 +134,15 @@ class FlacUpgradeWorker @AssistedInject constructor(
         const val KEY_UPGRADED = "flac_upgraded"
         const val KEY_NO_MATCH = "flac_no_match"
         const val KEY_FAILED = "flac_failed"
+
+        /**
+         * Input-data flag set ONLY by TrackDownloadWorker's automatic
+         * post-sync sweep. Marks the batch as machine-initiated so the
+         * consent re-check in [doWork] can discard it when the master
+         * lossless toggle is off — while user-enqueued batches
+         * (FlacUpgradeEnqueuer, untagged) run regardless.
+         */
+        const val KEY_AUTO_SWEEP = "flac_auto_sweep"
         private const val TAG = "FlacUpgradeWorker"
     }
 }
--- a/core/data/src/main/kotlin/com/stash/core/data/sync/workers/TrackDownloadWorker.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/sync/workers/TrackDownloadWorker.kt
@@ -66,6 +66,7 @@ class TrackDownloadWorker @AssistedInject constructor(
     private val fileAdopter: com.stash.core.data.library.FileAdopter,
     private val syncLog: com.stash.core.data.sync.SyncLog,
     private val flacUpgradeQueueDao: com.stash.core.data.db.dao.FlacUpgradeQueueDao,
+    private val losslessUpgrader: com.stash.core.data.lossless.LosslessUpgrader,
 ) : CoroutineWorker(appContext, params) {
 
     companion object {
@@ -244,6 +245,7 @@ class TrackDownloadWorker @AssistedInject constructor(
             val totalBytesDownloaded = AtomicLong(0)
             val firstError = AtomicReference<String?>(null)
             val playlistsChecked = inputData.getInt(DiffWorker.KEY_PLAYLISTS_CHECKED, 0)
+            val downloadedTrackIds = java.util.Collections.synchronizedSet(mutableSetOf<Long>())
 
             supervisorScope {
                 for (queueItem in pendingItems) {
@@ -390,6 +392,7 @@ class TrackDownloadWorker @AssistedInject constructor(
                                     )
                                     totalBytesDownloaded.addAndGet(fileSize)
                                     downloadedCount.incrementAndGet()
+                                    downloadedTrackIds.add(queueItem.trackId)
                                 }
                                 is TrackDownloadOutcome.Unmatched -> {
                                     val err = "No YouTube match for: ${track.artist} - ${track.title}"
@@ -602,7 +605,7 @@ class TrackDownloadWorker @AssistedInject constructor(
                 )
             }
 
-            runLosslessUpgradeSweep()
+            runLosslessUpgradeSweep(downloadedTrackIds)
 
             return Result.success(
                 workDataOf(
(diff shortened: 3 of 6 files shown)


rawnaldclark (changes requested):
Thanks for this — and especially for splitting it into two independent fixes with a clear write-up of each. The first one I'd take almost as-is; the second needs rework before it can land.

**Fix 1 (gating the upgrade sweep) — this is right, and the placement is the good part.** You gated at the sweep call-site in `TrackDownloadWorker` rather than inside `upgradeToLossless`, which keeps the Now Playing manual "upgrade to lossless" button working — that one deliberately relies on `forced = true` bypassing the master toggle (see `DownloadManagerDeferTest.kt:206`). Easy to get wrong; you got it right. The `downloadedTrackIds` filter is also correct: `synchronizedSet` covers the concurrent adds from the `launch` children, the read happens after `supervisorScope` joins, and being per-run means a track that landed lossy this sync is properly re-eligible next sync.

Two things on it:
- `LosslessSourcePreferences` already has `enabledNow()`, which is exactly `enabled.first()` and is what `DownloadManager.kt:217` / `SearchDownloadCoordinator.kt:182` use. Please delegate to it rather than re-implementing, and drop the now-unused `flow.first` import. (Also `LosslessUpgraderImpl.kt:13` imports `LosslessSourcePreferences` from its own package — that one can just go.)
- The gate only covers *enqueueing*. `FlacUpgradeWorker.doWork()` never checks it, and `upgradeToLossless` uses `forced = true`, so a batch enqueued before the user flips the toggle off still downloads to completion — `setEnabled(false)` requeues WAITING_FOR_LOSSLESS rows but doesn't clear `flac_upgrade_queue`. Could you add the check at the top of `FlacUpgradeWorker.doWork()` (`clearPending()` + `Result.success()`)? That also gives the new `isLosslessEnabled()` a caller on the worker whose test you had to stub it into.

**Fix 2 (`resolveInternalTargetFile`) — I can't take this one as written; it turns a recoverable bug into a destructive one.** The scan at `FileOrganizer.kt:359-367` matches on the title slug alone, with no check that the file it finds belongs to the track being committed, and `commitDownload:320` then copies over it with `overwrite = true`. Two different tracks by the same artist sharing a title slug is common — `Intro`/`Outro`/`Interlude`, an album cut plus the greatest-hits version, a deluxe reissue. Concretely: `/music/kanye-west/the-college-dropout/intro.m4a` exists; a sync downloads "Kanye West – Intro" from *Late Registration*; the scan returns the College Dropout file, the new audio overwrites it, and both rows now carry the same `file_path`. The old bug orphaned a file (wasteful, recoverable); this one destroys audio and cross-wires two rows. Deleting either track then takes the other's file with it.

Three more things in the same helper:
- There's no extension filter, so `.lrc` lyrics sidecars match too — `LyricsSidecarWriter` writes `<basename>.lrc` into that same album folder, so `creep.lrc` matches `titleSlug == "creep"` exactly as well as `creep.flac`, and `listFiles()` order picks the winner. When the sidecar wins we write audio into the `.lrc` and set `file_path` to it.
- The returned file keeps its *old* extension and the `format` argument is dropped, so a lossless upgrade writes FLAC bytes into `title.opus`. `file_format` still comes out right (`AudioDurationExtractor` sniffs the MIME rather than the extension), but `walkInternalDir` classifies lossless by extension, so Home's storage numbers drift — and any other player or tagger sees a `.opus` that isn't one, which cuts against the "library you can take anywhere" promise in the class KDoc.
- `listFiles()` has no defined order, so when several album folders match, which one wins is arbitrary and can change between runs.

Also worth flagging that the blast radius is bigger than the sync path: `commitDownload` also backs `TrackFinalizer` (every lossless download), `LocalImportCoordinator`, `SwapCoordinator`, and `FailedMatchesViewModel`. The local-import case is the one that worries me most — importing a file would silently overwrite an existing downloaded track, then insert a second row pointing at it.

On the underlying problem — I think the diagnosis may be aimed slightly off. The KDoc argues the old path can't be threaded through as a parameter, but `DownloadManager` re-reads the row at `:275` and already deletes the orphan at `:350-359` when the paths differ, and `LosslessUpgraderImpl.persistUpgrade` does the same at `:130-132`. If you have a repro where a file is genuinely orphaned, I'd love to see which commit path it goes through — my guess is it's a caller that's missing that cleanup step, and adding it there would be a much smaller change than rediscovering the target from disk. If we do end up needing a disk scan, it should at minimum be restricted to known audio extensions, require a unique match, and preserve the new format's extension.

Last thing: no test covers the new helper (there's no `FileOrganizer` commit test at all) or the new gate — both test edits are fixture updates. Whatever shape fix 2 lands in, a `TemporaryFolder`-based test for the same-slug-different-album case would be worth having.

**Coordination:** #433 carries all six of these files with the same changes, so it's effectively stacked on this one. Let's land #430 first once fix 2 is sorted, then rebase #433 onto master and drop the duplicated hunks — note `FileOrganizer.kt` has already drifted between the two branches (+44/−1 here vs +43/−1 there), so that rebase wants a real read rather than a blind resolution.

Verified locally on your head commit: `:core:data:testDebugUnitTest` 666 passed / 0 failed, `:data:download:testDebugUnitTest` 595 passed / 0 failed / 5 skipped, `:core:data:compileDebugKotlin` clean. Green, matching the master baseline — the issues above are all logic ones the current suites don't reach.

One note on the linked issue: this doesn't close the "upgrade sweep re-batches every lossy track each sync" follow-up from #413 — that one needs a persisted last-attempt stamp, whereas this filter is in-memory and single-run. Still a genuine improvement for the within-run double-attempt, just a different slice.

ParaliyzedEvo:
@rawnaldclark should be good to go now thanks for @JoshDui

rawnaldclark (changes requested):
Thanks for the quick turnaround — the fix-1 cleanups are all exactly right. `enabledNow()` delegation is in at `LosslessUpgraderImpl.kt:49`, the stray imports are gone, and `FlacUpgradeWorker.doWork()` now gates with `clearPending()`, which is the right DAO call (PENDING only, terminal rows preserved). I also like that you reworked `passes forced = true` into a real assertion — `coVerify(exactly = 0) { losslessPrefs.enabledNow() }` is a proper regression guard for the manual override, better than what was there before. And dropping fix 2 rather than patching around it was the right call.

One problem, and it's mine as much as yours: the `doWork()` gate I asked for lands on a worker with two producers.

`FlacUpgradeWorker` is driven by `TrackDownloadWorker`'s auto-sweep *and* by `FlacUpgradeEnqueuer` (`FlacUpgradeEnqueuer.kt:28`), which is the user-initiated **"Upgrade to FLAC"** multi-select action — `LibraryScreen.kt:252` → confirm dialog → `LibraryViewModel.upgradeSelectedToFlac():636`. Nothing in that path consults the master toggle.

So: a user with "Lossless downloads" off (which is exactly the user who upgrades selectively) picks 12 tracks and taps Upgrade. `LibraryViewModel.kt:640` promises *"Upgrading 12 songs to FLAC — watch the notification."* The worker starts, sees the toggle off, clears the queue, returns success. No notification, no error, nothing. Meanwhile the Now Playing single-track upgrade still works with the toggle off, because it calls `upgradeToLossless` directly with `forced = true` — the manual-override contract you correctly protected in fix 1. The batch sibling now silently drops it.

Worth noting the race the gate was meant to close is actually widest on the *wrong* path: `FlacUpgradeEnqueuer` sets wifiOnly constraints so its request can sit queued for hours, while the sweep's request (`TrackDownloadWorker.kt:869`) has no constraints and starts almost immediately.

Smallest fix is to tag the automatic batch and gate only that — `setInputData(workDataOf(KEY_AUTO_SWEEP to true))` on the sweep's request, and `if (inputData.getBoolean(KEY_AUTO_SWEEP, false) && !isLosslessEnabled())` at the top of `doWork()`. Your new test just needs the flag set.

Two smaller things:

- The gate only fires at worker start, so a batch already draining keeps going after the toggle flips — and per the class KDoc these run for hours. If you want that closed too, `LosslessSourcePreferences.setEnabled(false)` already has a side-effect block next to `requeueWaitingForLossless()`; a `clearAll()` + `cancelUniqueWork(FlacUpgradeWorker.UNIQUE_WORK_NAME)` there handles both windows at the source.
- Could you trim the PR body? It still describes `resolveInternalTargetFile` and lists `FileOrganizer.kt`, and still says `Closes #219`. #230 is genuinely fixed by the gate, but #219 ("deletes the non-flac files") isn't touched by this diff — `persistUpgrade` already deletes the old file only after a successful write. The body drives the release notes, so I'd rather it match the code.

Also flagging for coordination: `resolveInternalTargetFile` is still in **#433** unchanged (`FileOrganizer.kt +43/−1`) with all four issues from last time. Dropping it here doesn't retire it — it'll want the same treatment there. #433 also adds `isLosslessEnabled()` to `LosslessUpgrader` with different KDoc and placement, so expect a conflict on that file when you rebase.

Verified locally on `8258e890`: `:core:data:testDebugUnitTest` 670 passed / 0 failed, `:data:download:testDebugUnitTest` 596 passed / 0 failed / 5 skipped, `:core:data:compileDebugKotlin` clean. Trial-merged master cleanly with `:app:compileDebugKotlin` green. All the new tests pass — the batch-action issue is a wiring one the suites don't reach, since there's no test that drives `FlacUpgradeWorker` through `FlacUpgradeEnqueuer`.

PR #433, opened 2026-08-16
Add download management screen

### LINKED ISSUE #224: Download management tab
I would love if the app could get a download tab where you get to see your download history and current downloading songs and queue, get to cancel a download and so on

ParaliyzedEvo:
## Summary
- add an always-visible Downloads entry from the Sync screen
- show durable active queue and recent download history with live per-track progress
- support retry and race-safe per-item cancellation for sync and discovery downloads
- add Room and ViewModel regression coverage for ordering, transitions, retries, and cancellation races

## Testing
- :core:data:testDebugUnitTest --tests com.stash.core.data.db.dao.DownloadQueueDaoManagementTest
- :feature:sync:testDebugUnitTest --tests com.stash.feature.sync.DownloadManagementViewModelTest
- compileDebugUnitTestKotlin
- :app:assembleDebug

## Scope
Manual Search/Artist/Album downloads currently use a separate in-memory pipeline and are intentionally left for a follow-up requiring durable origin tracking.

Closes rawnaldclark/Stash#224

Original PR by @JoshDui

### DIFF
--- a/app/src/main/kotlin/com/stash/app/navigation/StashNavHost.kt
+++ b/app/src/main/kotlin/com/stash/app/navigation/StashNavHost.kt
@@ -28,6 +28,7 @@ import com.stash.feature.settings.SettingsHubScreen
 import com.stash.feature.settings.equalizer.EqualizerScreen
 import com.stash.feature.settings.libraryhealth.LibraryHealthScreen
 import com.stash.feature.sync.FailedDownloadsScreen
+import com.stash.feature.sync.DownloadManagementScreen
 import com.stash.feature.sync.FailedMatchesScreen
 import com.stash.feature.sync.ManagePlaylistsScreen
 import com.stash.feature.sync.SyncScreen
@@ -179,6 +180,9 @@ fun StashNavHost(
                 onNavigateToBlockedSongs = { navController.navigate(BlockedSongsRoute) },
                 onNavigateToFailedDownloads = {
                     navController.navigate(FailedDownloadsRoute)
+                },
+                onNavigateToDownloads = {
+                    navController.navigate(DownloadManagementRoute)
                 },
                 onNavigateToSettings = {
                     navController.navigate(SettingsRoute) {
@@ -360,6 +364,9 @@ fun StashNavHost(
 
         composable<FailedDownloadsRoute> {
             FailedDownloadsScreen(onBack = { navController.popBackStack() })
+        }
+        composable<DownloadManagementRoute> {
+            DownloadManagementScreen(onBack = { navController.popBackStack() })
         }
 
         composable<SearchArtistRoute> {
--- a/app/src/main/kotlin/com/stash/app/navigation/TopLevelDestination.kt
+++ b/app/src/main/kotlin/com/stash/app/navigation/TopLevelDestination.kt
@@ -33,6 +33,7 @@ enum class TopLevelDestination(val selectedIcon: ImageVector, val unselectedIcon
 @Serializable data class LikedSongsDetailRoute(val source: String? = null)
 @Serializable data object FailedMatchesRoute
 @Serializable data object FailedDownloadsRoute
+@Serializable data object DownloadManagementRoute
 @Serializable data object BlockedSongsRoute
 @Serializable data object EqualizerRoute
 @Serializable data object LibraryHealthRoute
--- a/core/data/src/main/kotlin/com/stash/core/data/db/dao/DownloadQueueDao.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/db/dao/DownloadQueueDao.kt
@@ -24,6 +24,22 @@ data class SourceCount(
     val cnt: Int,
 )
 
+/** Joined queue/track row used by the download-management screen. */
+data class DownloadManagementRow(
+    val queueId: Long,
+    val trackId: Long,
+    val title: String,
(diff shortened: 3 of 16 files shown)


rawnaldclark (changes requested):
Thanks for picking this up, @ParaliyzedEvo — and credit to @JoshDui for the original PR this is adopted from. #224 has been open a while and the shape you've landed on is the right one: moving the queue onto compare-and-swap DAO writes instead of unconditional `updateStatus` calls is a real improvement over what was there, and `DownloadQueueDaoManagementTest` is genuinely excellent — a real in-memory Room DB exercising the actual transition semantics, with `cancel_then_late_terminal_writes_are_noops` and `completion_then_cancel_preserves_completion` covering exactly the races that matter. That's the standard I'd like the rest of the suite to meet.

A few things need addressing before this can go in.

**The `:core:data` suite is red.** `DiscoveryDownloadWorkerTest` goes 12/12 green on master and 4/12 on this branch. The cause is mechanical: `downloadQueueDao` is `mockk(relaxed = true)`, so the new `claimForDownload()` returns `0`, `processQueueItem` returns on its first line, and the worker body never executes — the mock traces show only `pendingDiscoveryDownloads` was ever called. One stub in `setUp` should fix it: `coEvery { downloadQueueDao.claimForDownload(any()) } returns 1`. Worth running the full `:core:data:testDebugUnitTest` and `:feature:sync:testDebugUnitTest` rather than just the two new classes.

**Cancellation isn't durable yet.** `SKIPPED` isn't recognised by anything else in `DownloadQueueDao`. `getUnqueuedTrackIds` excludes only `PENDING/IN_PROGRESS/FAILED`, so on the next sync `LibraryReconciliationUseCase` sees a cancelled track as unqueued and inserts a fresh `PENDING` row — the download restarts. And because no cleanup query matches `SKIPPED`, the original row is never swept, so History accumulates a permanent duplicate per cancel. Adding `SKIPPED` to the requeue exclusion and to `deleteOrphanedQueueEntries` should cover both.

**Chain-mode cancel still completes the download.** The `Success` branch in `TrackDownloadWorker.doWork()` (~L314) doesn't re-check `SKIPPED` or delete the file, unlike single-track mode (L762) and the discovery worker. Cancel a track that finishes in the race window and `markAsDownloaded` still runs — the track lands in the library while Downloads reports "Cancelled". Mirroring the single-track guard would close it.

**The cancellation guards need `NonCancellable`.** In `catch (ce: CancellationException) { if (dao.getStatusById(...) == SKIPPED) ... }`, `getStatusById` is a Room suspend query, and `withContext` throws on an already-cancelled Job before reaching the DB. When `DownloadJobRegistry.cancel()` cancelled that coroutine — the case these guards exist for — the check never runs, taking the file cleanup with it. `withContext(NonCancellable) { ... }` around those reads should do it. A test that cancels a real job rather than asserting on mocks would be worth adding here.

**`observeActiveDownloadRows()` needs `@Transaction` and a bound.** It's an unbounded 12-column join observed live while the sync worker rewrites `download_queue` on every status transition — the same shape as the #380 crash (`Couldn't read row 2430 from CursorWindow`). See the notes on `TrackDao.getPlaylistTracks` / the album-group query for the pattern.

**On the `FileOrganizer` change** — I see this is coming from #430, since this branch is stacked on it. Two things concern me there, and I've left the full detail on that PR: `resolveInternalTargetFile` returns the existing file *with its current extension*, ignoring the `format` argument, so a FLAC upgrade over an `.opus` file writes FLAC bytes into a `.opus` path; and matching on title slug across every album folder means a single and its album version (or a studio and live take) resolve to the same file, with the second download overwriting the first.

**Stacking.** Since `bbd44aef` here is tree-identical to #430's head, merging this as-is would land #430 unreviewed. Could we get #430 in first and then rebase this so the diff is only the download-management work? Happy to review both in that order.

Smaller things: `WAITING_FOR_LOSSLESS` rows render in their own section but get no cancel button (`toItem`'s action mapping only covers `PENDING`/`IN_PROGRESS`), though `markCancelledIfActive` supports them. `state.queued.indexOf(item)` inside the item lambda is O(n²) — `itemsIndexed` would avoid it. The progress `combine` re-maps all 100 history rows on every byte callback; conflating progress and keeping history out of that combine would help. And `core/ui/components` already has `GlassCard`, `SectionHeader` and `CrispChipRow` for the patterns hand-rolled in the new screen.

Navigation, schema and the `CancellationException` re-throw discipline all check out — no migration needed, and it merges cleanly onto current master with `:app:compileDebugKotlin` green.

rawnaldclark:
For transparency on the final shape: I finished this one off maintainer-side rather than doing another review round. The branch was rebased onto current master (which dropped the old #430-era base — that's why the diff shrank to just the download-management work), and the open review items landed in one fix commit on top of @JoshDui's feature commit: the DiscoveryDownloadWorkerTest stubs/assertions, durable cancel (SKIPPED in the reconciliation + orphan predicates, lockstep with the diagnostic count), retry-from-History for cancelled rows, the chain-mode completion guard, NonCancellable around the six cancellation-guard reads, @Transaction + LIMIT on the active-rows observer, and the itemsIndexed/history-combine perf touches. Full suites green (692 core:data / 45 sync). Thanks to you both — the compare-and-swap DAO design and its test class carried this over the line.

PR #439, opened 2026-08-19
feat(library): official artist photos

### LINKED ISSUE #388: [Bug]: queue shaking when its exacly first out of 10 tracks

yuriikot2010:
## What

Gives the Library Artists tab real artist photos instead of album-art proxies / gradient initials, and the Artist Detail header the same official avatar.

- **Backfill worker** (`ArtistImageBackfillWorker`) walks the distinct artist credits in the library, collapses them to their primary act, resolves each via a new tri-state `resolveArtistPhoto` API, and caches the result in a new `artist_images` table (DB v43, migration 42 -> 43).
- **Sentinel semantics**: only a genuine "no avatar" answer stamps the permanent sentinel; transient failures (network / rate limit) write nothing and are retried on the next pass — mirrors `ArtBackfillWorker`.
- **Parser**: `core:common` gains `String.primaryArtist()` (comma-only split + allowlist) so collaboration credits like "Aarne, Toxi$, Big Baby Tape" display as "Aarne" without shredding band names that merely contain separators.
- **UI**: `ArtistAvatar` in the Artists grid (photo -> album-art proxy -> gradient initial), `LibraryViewModel` regroups by primary artist and overlays photos, `ArtistDetailScreen` header shows the cached photo with the `Person`-icon placeholder fallback.
- **Batch & case fixes** per review: single transactional upsert per pass, targeted `observeByName` query, COLLATE NOCASE key, `CancellationException` rethrown, `remaining=` clamped to 0.

This PR is the artist-photos slice of the feature. The album-hero FlowRow wrap (#385, PR #441), the album-card tap navigation (#386, PR #442), and the display/match credit-parsing work (PR #440) live in their own PRs. Does **not** fix #388 (the artist-text list-scroll shuffle is a separate `.animateItem()` drag-reorder follow-up).

### DIFF
--- a/app/src/main/kotlin/com/stash/app/StashApplication.kt
+++ b/app/src/main/kotlin/com/stash/app/StashApplication.kt
@@ -28,6 +28,7 @@ import com.stash.core.data.sync.SyncNotificationManager
 import com.stash.data.download.backfill.MetadataBackfillScheduler
 import com.stash.data.download.ytdlp.YtDlpManager
 import com.stash.core.data.sync.workers.ArtBackfillWorker
+import com.stash.core.data.sync.workers.ArtistImageBackfillWorker
 import com.stash.core.data.sync.workers.AutoSaveScrobbler
 import com.stash.core.data.sync.workers.DiscoveryDownloadWorker
 import com.stash.core.data.sync.workers.QualityInfoBackfillWorker
@@ -442,6 +443,11 @@ class StashApplication : Application(), Configuration.Provider {
         // no thumbnail (see ArtBackfillWorker KDoc). KEEP policy means the
         // worker is a no-op on every subsequent launch once it completes.
         ArtBackfillWorker.enqueueOneTime(this)
+        // First-launch artist-photo backfill: resolves the official avatar
+        // for every artist on the Library Artists tab (see ArtistImageBackfillWorker
+        // KDoc). KEEP policy means a re-launch before the worker completes
+        // doesn't re-enqueue; re-arms after each sync via SyncFinalizeWorker.
+        ArtistImageBackfillWorker.enqueueOneTime(this)
         // v0.9.11: kick a background sweep that fills in bit-depth +
         // sample-rate for tracks downloaded before the columns existed.
         // The flag is set immediately on enqueue (not after success) so
--- a/core/data/schemas/com.stash.core.data.db.StashDatabase/44.json
+++ b/core/data/schemas/com.stash.core.data.db.StashDatabase/44.json
@@ -1,0 +1,2168 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 44,
+    "identityHash": "b97f8e7a010533ebaa7367e4272f4e2a",
+    "entities": [
+      {
+        "tableName": "tracks",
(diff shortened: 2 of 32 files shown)


rawnaldclark:
Heads-up: I've rebased this branch onto current master and force-pushed it here (maintainer edits were on). Your two commits are unchanged and still yours — `97bbea1d` and `1d4350fb`. **If you have local work on `pr3-artist-photos`, `git fetch` and reset before committing anything else**, or you'll fight the rewrite.

Why: #440 merged first, and both PRs added `core/common/ArtistCredits.kt`, so this went `CONFLICTING` the moment it landed. It's `MERGEABLE` again now.

What the conflicts turned out to be — all mechanical, nothing behavioural:

- **`ArtistCredits.kt` / `ArtistCreditsTest.kt`** — took master's copy. #440's version is a strict superset: same `ARTIST_CREDIT_ALLOWLIST`, same `primaryArtist()`, same `isAllowlistedCredit()`, plus `splitArtistCredits`/`expandArtistCredits`/`matchesArtistCredits`. Nothing yours was dropped, it's just already there.
- **`core/common/build.gradle.kts`** — kept master's comment; the junit dep is the same line either way.
- **`LibraryViewModel.kt`** — import union, `matchesArtistCredits` + `primaryArtist` + `ArtistImageDao`.
- **`LibraryScreen.kt`** — your side was a pure duplicate; all three coil3 imports already existed on master.

The photo overlay and #440's credit matching sit in different parts of those two files, so everything else merged automatically.

Verified after the rebase: `:core:common`, `:core:data` and `:feature:library` all compile and pass, including `MigrationV42V43Test` and the `ArtistImageBackfillWorker` tests.

**One thing left before merge, and I left it for you deliberately:** `core/data/schemas/.../43.json` looks hand-written rather than Room-generated. Its `createSql` for `artist_images` omits `COLLATE NOCASE` even though the `@Entity` and `MIGRATION_42_43` both declare it, and it carries `VALUES(42, ...)` in `setupQueries` inside a version-43 file. Not a runtime bug — both the create path and the migrate path build the column correctly, and no CI job gates on schema drift — but a clean build will regenerate it with a different `identityHash` and leave the schema dir dirty. Could you regenerate it from the entity (run the Room schema export for `:core:data`) and push that? Hand-patching it is what produced the mismatch, so I didn't want to hand-patch it again.

Worth saying: the feature itself is in good shape. The tri-state Resolved/NoAvatar/Failed sentinel, the cancellation handling and the backfill worker's retry semantics all read well, and it's properly tested. Two small things you can take or leave: the intra-batch dedup is case-sensitive (`.distinct()` on original-case names, so `Aarne` and `AARNE` each cost a search before the NOCASE key collapses them — `.distinctBy { it.lowercase() }` fixes it), and `setBackoffCriteria(EXPONENTIAL, 30s)` is unreachable because every path returns `Result.success()`.

Thanks for this one — and sorry for the rebase churn, that's the cost of two PRs landing on the same new file.

rawnaldclark:
Queued after the smaller ones; it's the largest of the open set and carries the v43 migration plus a new backfill worker that runs at first launch and after every sync. Before merge I want one measured pass on a real library (~860 artists here): how many InnerTube calls the first backfill makes and what it costs on battery, since it runs unmetered-only but on every sync completion. Migration ordering versus #452 as above.

rawnaldclark:
Heads-up after today's master push: master now carries a schema migration to **version 43** (`MIGRATION_42_43`, data-only, from `2b0be6b5`, with `43.json` committed). This PR's `artist_images` migration and `MigrationV42V43Test` therefore collide on both the version number and the test class name.

To land it: rebase onto master, renumber to `MIGRATION_43_44` (database `version = 44`), rename the test to `MigrationV43V44Test`, and regenerate `44.json` from a build rather than by hand (the earlier review noted the hand-edited schema was missing `COLLATE NOCASE` and had a copy-paste `VALUES(42, …)`). PR #452 is adding an `ALL_MIGRATIONS` chain test, so once both are in, ordering mistakes show up as a red build instead of a crash on upgrade.

rawnaldclark (changes requested):
Read the schema, the DAO, the backfill worker, the resolver and the Library wiring after your rebase to 44. The migration is right (NOCASE key, `attempted_at NOT NULL` matching the entity default, `44.json` committed, chain test renamed), the tri-state resolution keeps transient failures out of the sentinel, cancellation is rethrown, and the batch upsert is one transaction. Good work, and thank you for the renumber.

Two things before it merges:

1. **Master moved under you again, in your favour this time.** #452 landed tonight and replaced the hand-maintained `addMigrations(...)` list with `StashDatabase.ALL_MIGRATIONS`, a lazy array, guarded by `AllMigrationsChainTest`, which fails the build when a `MIGRATION_*` is missing from it. Rebase, add `MIGRATION_43_44` to `ALL_MIGRATIONS`, and drop the `DatabaseModule` line; the test will tell you if anything else is off.
2. **Streaming-only libraries get no photos.** `distinctArtistNames()` selects `is_downloaded = 1` only. Most Online-mode users have few or no downloads, so their Artists tab keeps the gradient initials while the feature ships "on". Widen it to artists of tracks that are downloaded or streamable, or of any live playlist membership, whichever matches what the Artists tab itself lists.

Not in tonight's v0.9.104 (it tagged before this review); with those two in, it's a device run on the test rig and a merge for the next one. Two smaller notes for the same push: `enqueueAfterSync` uses `REPLACE`, so a sync that finishes mid-backfill restarts it from the top; `KEEP` would let a running pass finish. And 150 artists per pass at 400 ms is a fine ceiling, but say in the KDoc that a big library fills over several syncs, so nobody reads a half-filled grid as a bug.

ParaliyzedEvo:
@rawnaldclark looks like the changes have been made

rawnaldclark:
Pushed one commit on top of yours (d01523e9) — maintainer edits, nothing of yours rewritten.

**What changed:** the Artists tab stays downloads-only (like Songs and Albums) and no longer follows the Online/Offline switch. Two reasons: since v0.9.107 that switch only decides what Sync writes to disk, never what a screen shows; and `is_streamable` is a column nothing in the app ever writes, so the widening changed nothing on real data — while `distinctArtistNames()` was paying an InnerTube search for every artist in liked songs and playlists that the tab never lists (~1,500 on a large library). `distinctArtistNames()` is now exactly the tab's set (`getAllArtists(includeStreamable = false)`), and the DAO test is six cases for that. Worker, resolver, UI and migration are yours as written.

Whether the Library should show stream-only content at all is a separate product decision; if it goes that way, both queries widen together and your membership query is the template.

Merging once CI is green. Thank you for the renumber and the follow-through on every round.

PR #415, opened 2026-08-09
Skip home items without owner data to prevent mislabeling

ParaliyzedEvo:
## Summary
- `parseHomeFeedForSpotifyMixes` previously defaulted `ownerId` to `"spotify"` whenever a home feed item's `ownerV2` data was missing. That default was meant to catch editorial mixes (Daily Mix, Discover Weekly, Made-For-You, "This Is <artist>") that legitimately have no owner, but it also swept in user-owned playlists from non-mix home sections (e.g. "Jump back in") whose cards simply omit `ownerV2` for other reasons — causing those to be mislabeled as `DAILY_MIX`.
- Fix: fail closed. If `ownerId` can't be resolved from `ownerV2.username` or `ownerV2.id`, skip the item entirely (`continue`) instead of assuming it's Spotify-owned. A debug log records the skip with the item's `name` for traceability.
- Updated the surrounding doc comment to describe the new skip behavior instead of the old default-to-`"spotify"` behavior.

## Changes
- `data/spotify/src/main/kotlin/com/stash/data/spotify/SpotifyApiClient.kt`
  - Removed `?: "spotify"` fallback when resolving `ownerId`
  - Added early `continue` (with debug log) when `ownerId` is null
  - Updated kdoc above `isSpotifyMix` usage to reflect fail-closed behavior

## Test plan
- [x] `./gradlew :data:spotify:compileDebugKotlin`
- [x] `./gradlew :data:spotify:test`
- [x] Device / Android version tested: S26u/Android 16

Closes https://github.com/ParaliyzedEvo/Stash/issues/124

PR #447, opened 2026-08-23
fix(sync): stop filing a user's Spotify playlists as mixes (#437)

### LINKED ISSUE #437: [Bug]: My Spotify play lists are not showing despite I have synchronised and I have downloaded the latest version
### Before you start

### Stash version

0.9.98

### Phone & Android version

Pixel 8 Pro

### Where does it happen?

It does not show my Spotify playlists. I have 5 playlists and it is only showing one. 

### Steps to reproduce

 Click synchronise, says it has found 5 playlists, goes through it quickly then it shows only one. I have attached screenshots. 

Image
Image
Image

chaudhary-lakshay:
Fixes #437 — "5 playlists found, 1 shown".

## Root cause

Three defects stacked up:

**1. A fabricated owner.** `parseHomeFeedForSpotifyMixes` defaulted `ownerId` to `"spotify"` when a home-feed item carried no `ownerV2` block. Spotify's home feed surfaces the user's OWN playlists under "Recently played" / "Jump back in", and those cards often omit `ownerV2` — so they took `isSpotifyMix`'s owner catch-all and were snapshotted as `DAILY_MIX`. Unknown owner is now fail-closed: no owner data means we can't confirm it's a real mix, so the item is skipped rather than mislabelled.

**2. `playlists.type` was write-once.** Whichever fetch pass saw a `source_id` first owned its type forever. A row misfiled as `DAILY_MIX` never returned to the Library Playlists tab or the Sync tab's "n/n PLAYLISTS" count — both filter on `type`. `findOrCreatePlaylist` now re-types a `DAILY_MIX` row that the library walk reports as `CUSTOM`. Deliberately one-way, so the type can't flap with the auto-mix-discovery toggle; `mix_number` is cleared alongside it.

**3. Deactivation measured the wrong list.** `deactivateMissingSpotifyCustomPlaylists` compared against the post-`keepAsLibraryPlaylist` list. That filter is classification routing, not evidence a playlist left the account — an id landing in `homeFeedMixIds`, or a name turning into "Daily Mix N", hid the `CUSTOM` row the walk had just listed. It now measures against everything the run observed.

## Verification

Unit: `ReconciledPlaylistTypeTest` (one-way rule, local-only types and `LIKED_SONGS` left alone), `SpotifyLivenessIdsTest`.

On-device, Samsung Galaxy Tab Active3 (SM-T570, Android 13):

- One real sync skipped 17 owner-less home-feed items that would previously have become bogus mixes (`calm vibes`, `digital detox✌️`, `sad instrumentals`, `Peaceful Piano`, …).
- Reproduced #437 by flipping a real playlist row to `type='DAILY_MIX', mix_number=7`. The Spotify card then read **1 MIXES · 0/1 PLAYLISTS** with two playlists in the account — the reported shape.
- One Sync Now healed it: `DiffWorker: Re-typed 'Stash' DAILY_MIX -> CUSTOM (source_id=0NIipN6vyz2yLYguJwbUC6)`. Card went to **0/2 PLAYLISTS**, MIXES cell gone, `mix_number` cleared. A second, untouched playlist row stayed `CUSTOM` as a control.
- Sync log surfaces it to the user: `Stash is your playlist, not a mix — moved to Playlists`.
- No crash; Room opened at schema v42 (the change adds a `@Query` only, no schema change, no migration).

Defect 3 is unit-covered only — reaching it on-device needs a playlist to disappear from the account mid-run.

## Known limit

While auto-mix discovery is on, `keepAsLibraryPlaylist` withholds every id the home feed claimed, so a Spotify-OWNED playlist the user saved stays `DAILY_MIX` — which is what it is. A user's own playlist is not claimed by the home feed any more (defect 1), so it reaches the library walk and heals, which is the #437 case. Documented in `reconciledPlaylistType`'s KDoc.

### DIFF
--- a/core/data/src/main/kotlin/com/stash/core/data/db/dao/PlaylistDao.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/db/dao/PlaylistDao.kt
@@ -658,6 +658,20 @@ interface PlaylistDao {
     @Query("UPDATE playlists SET is_active = 1 WHERE id = :playlistId AND is_active = 0")
     suspend fun reactivateById(playlistId: Long): Int
 
+    /**
+     * Re-file a playlist under the type today's snapshot reports.
+     *
+     * `type` used to be write-once — whichever fetch pass saw the `source_id`
+     * first owned it forever. A playlist the Spotify home-feed mix pass had
+     * filed as DAILY_MIX therefore never returned to the Library Playlists
+     * tab, no matter how many syncs saw it as the user's own (issue #437).
+     *
+     * [mixNumber] travels with the type: it only means anything for a mix, so
+     * a row leaving DAILY_MIX must not keep a stale ordinal.
+     */
+    @Query("UPDATE playlists SET type = :type, mix_number = :mixNumber WHERE id = :playlistId")
+    suspend fun updateType(playlistId: Long, type: PlaylistType, mixNumber: Int?)
+
     // ── Custom playlist management ──────────────────────────────────────
 
     /** Get the next available position for appending a track to a playlist. */
--- a/core/data/src/main/kotlin/com/stash/core/data/sync/workers/DiffWorker.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/sync/workers/DiffWorker.kt
@@ -28,6 +28,7 @@ import com.stash.core.model.MusicSource
 import com.stash.core.model.PlaylistType
 import com.stash.core.model.SyncMode
 import com.stash.core.model.SyncState
+import com.stash.data.spotify.SpotifyApiClient
 import dagger.assisted.Assisted
 import dagger.assisted.AssistedInject
 import kotlinx.coroutines.flow.first
@@ -76,20 +77,50 @@ internal fun defaultSyncEnabled(type: PlaylistType, online: Boolean): Boolean = 
  * per-source singleton with a synthetic source id that no library walk should
  * ever re-type.
  *
- * Known limit: this only sees ids that reach a CUSTOM snapshot, and while
- * auto-mix discovery is on `keepAsLibraryPlaylist` withholds every id the home
- * feed claimed. A Spotify-OWNED playlist the user saved therefore stays
- * DAILY_MIX — which is what it is. A user's OWN playlist is not claimed by the
- * home feed (isSpotifyMix skips owner-less items as of v0.9.99), so it reaches
- * the library walk and heals here, which is the #437 case. Reuniting the two
- * would mean plumbing the library walk's id list into the diff pass; do that
- * only if a saved Made-For-You playlist ever needs to live under Playlists.
+ * Scoped to Spotify ids OUTSIDE Spotify's own namespace, because the type pair
+ * alone does not identify the #437 case and the filters that used to stand in
+ * for it are not load-bearing:
+ *
+ * - YouTube reaches the same pair with a real mix. A saved auto-mix tile is a
+ *   `VLRD…` browseId, and `parseSinglePlaylistFromTwoRowRenderer` accepts any
+ *   VL-prefixed id but `VLLM`/`VLSE` and strips the `VL` — yielding the very
+ *   `RD…` id the home-mix pass snapshotted as DAILY_MIX, now arriving again as
+ *   CUSTOM. `getPlaylistTracks` already documents that `getUserPlaylists()`
+ *   returns saved radios. Nothing on the YouTube path filters this.
+ * - On Spotify, `keepAsLibraryPlaylist` only withholds ids in `homeFeedMixIds`,
+ *   and that set is filled solely inside the home-feed `Success` branch. One
+ *   `Empty`, `Error`, thrown exception, or discovery-off run leaves it empty,
+ *   after which every saved mix whose name isn't literally "Daily Mix N" —
+ *   Discover Weekly, Release Radar, On Repeat, daylist, Blends, the yearly
+ *   recaps — is snapshotted CUSTOM.
+ *
+ * A wrong re-type is worse than the bug this fixes, and the rule is one-way so
+ * it never heals: the row leaves the Home mix rails, `mix_number` is gone,
+ * DiffWorker's syncEnabled gate exempts DAILY_MIX only so the row is skipped
+ * forever after, and `shouldEnqueueForDownload(CUSTOM, offline)` is true — so
+ * one "Enable all" queues an entire rotating mix, the #368 regression the
+ * surrounding code exists to prevent.
+ *
+ * [SPOTIFY_OWNED_ID_PREFIX] is the honest discriminator: Spotify generates
+ * every mix inside it, and the #437 playlists — the user's own — are never in
+ * it. So a Spotify-generated mix is protected whether or not the home-feed pass
+ * ran, and YouTube cannot reach the reconcile at all.
  */
 internal fun reconciledPlaylistType(
     existing: PlaylistType,
     snapshot: PlaylistType,
+    source: MusicSource,
+    sourceId: String,
 ): PlaylistType? =
-    if (existing == PlaylistType.DAILY_MIX && snapshot == PlaylistType.CUSTOM) snapshot else null
+    if (source == MusicSource.SPOTIFY &&
+        !sourceId.startsWith(SpotifyApiClient.SPOTIFY_OWNED_ID_PREFIX) &&
+        existing == PlaylistType.DAILY_MIX &&
+        snapshot == PlaylistType.CUSTOM
+    ) {
+        snapshot
+    } else {
+        null
+    }
 
 /**
  * Whether a playlist's tracks should be enqueued for download during this sync.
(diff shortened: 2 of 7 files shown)


rawnaldclark (changes requested):
Thanks for this — the write-up is genuinely excellent, and the diagnosis of #437 is right. Two notes before the main point, both in your favour:

- **Defect (a), the fabricated `ownerId`, already shipped on master** as `48e156da` (#415), so it isn't in this diff. No action needed, just flagging so the PR description doesn't confuse the next reader.
- **Defect (c), the `spotifyLivenessIds` widening, is correct and strictly safer** — it can only ever deactivate fewer rows than before. Happy with that as-is.

The blocker is defect (b), the type reconcile.

## `reconciledPlaylistType` is source-blind, and the guard that protects mixes isn't

`reconciledPlaylistType` (`DiffWorker.kt:88-95`) takes only the two enum values — no source, no source id — so it fires for **any** source and **any** id. The only thing shielding a genuine mix from being re-typed is `keepAsLibraryPlaylist`, which exists on the **Spotify path only** and is itself conditional on the home-feed pass having succeeded. That leaves two paths where a real algorithmic mix is permanently re-filed as `CUSTOM`:

**1. YouTube mixes — fires on the first sync after upgrade, no failure required.**

Home mixes are snapshotted `DAILY_MIX` with `sourcePlaylistId = mix.playlistId`, an `RD…` id. The library walk's `parseSinglePlaylistFromTwoRowRenderer` (`YTMusicApiClient.kt:1017-1027`) accepts any `VL`-prefixed browseId except `VLLM`/`VLSE` and does `browseId.removePrefix("VL")` — so a saved auto-mix tile (`VLRD…`: My Supermix, Discover Mix, New Release Mix, Replay Mix) yields **the identical `RD…` id**, which `fetchAndSnapshotUserPlaylist` snapshots as `CUSTOM`. Both reach `findOrCreatePlaylist` for the same `source_id`, and the CUSTOM one triggers `updateType(..., CUSTOM, mixNumber = null)`.

`getPlaylistTracks`'s own comment (`YTMusicApiClient.kt:212-215`) already states that `getUserPlaylists()` returns saved radios, so this collision is documented in the codebase.

**2. Spotify mixes that aren't literally named "Daily Mix N".**

`homeFeedMixIds` is only populated inside the `SyncResult.Success` branch (`PlaylistFetchWorker.kt:454-455`); `Empty` (:517), `Error` (:521), the outer catch (:528) and the discovery-off skip (:423-425) all leave it empty. `keepAsLibraryPlaylist` then returns true for every saved mix whose name doesn't match `DAILY_MIX_NAME = Regex("Daily Mix \\d+")` — which misses Discover Weekly, Release Radar, On Repeat, daylist, Blends and the yearly recaps. One transient `sp_dc` failure or a 429 is enough.

## Why this is worse than the bug it fixes

The rule is deliberately one-way, so **neither path heals**. Once a row flips to `CUSTOM`:

- it leaves the Home mix rails (`mixRail` returns null for non-`DAILY_MIX`),
- `mix_number` is discarded irrecoverably,
- `DiffWorker.kt:242` only exempts `DAILY_MIX` from the `syncEnabled` gate, so the row is skipped on every later run — its tracks freeze while the fetch worker keeps re-downloading its snapshot,
- and `shouldEnqueueForDownload(CUSTOM, offline)` is true, so the mis-filed mix now sits under "Your playlists" and one tap of **Enable all** queues every one of its tracks. That's the regression #368 that the surrounding code exists to prevent.

## Suggested fix — one guard, using a constant that already exists

Scope the reconcile to the case it was written for. `SPOTIFY_OWNED_ID_PREFIX = "37i9dQ"` already exists at `SpotifyApiClient.kt:146` for exactly this distinction, and the user-owned playlists from #437 never carry it:

```kotlin
internal fun reconciledPlaylistType(
    existing: PlaylistType,
    snapshot: PlaylistType,
    source: MusicSource,
    sourceId: String,
): PlaylistType? =
    if (source == MusicSource.SPOTIFY &&
        !sourceId.startsWith(SPOTIFY_OWNED_ID_PREFIX) &&
        existing == PlaylistType.DAILY_MIX &&
        snapshot == PlaylistType.CUSTOM
    ) snapshot else null
```

That closes both paths at once: YouTube can no longer reach the reconcile at all, and every Spotify-generated mix lives in the `37i9dQ…` namespace so it's protected regardless of whether the home-feed pass succeeded.

If you'd rather keep it run-scoped, the alternative is to plumb "the home-feed pass ran and succeeded this run" into `DiffWorker` (alongside `KEY_YOUTUBE_INVENTORY_COMPLETE`) and skip the reconcile when it's false — that also covers the discovery-off case you noted under "Known limit".

## One test ask

`ReconciledPlaylistTypeTest` is a pure-function truth table, and both bugs above **pass it** — the function is correct for the pairs it's handed; the defect is which pairs reach it. `DiffWorkerTest.kt` is already `@RunWith(RobolectricTestRunner)` with an in-memory Room DB, so the negative case is cheap to add there: seed a `DAILY_MIX` row, run a diff carrying a `CUSTOM` snapshot for the same `source_id`, and assert the type and `mix_number` survive.

Everything else here is solid — it's just this one guard. Happy to look again once it's in.

chaudhary-lakshay:
You're right on all three counts, and the blocker is a real one — thanks for taking the time to trace both paths.

Pushed `0675bf53` with your suggested guard, verbatim in shape:

```kotlin
internal fun reconciledPlaylistType(
    existing: PlaylistType,
    snapshot: PlaylistType,
    source: MusicSource,
    sourceId: String,
): PlaylistType? =
    if (source == MusicSource.SPOTIFY &&
        !sourceId.startsWith(SpotifyApiClient.SPOTIFY_OWNED_ID_PREFIX) &&
        existing == PlaylistType.DAILY_MIX &&
        snapshot == PlaylistType.CUSTOM
    ) snapshot else null
```

`SPOTIFY_OWNED_ID_PREFIX` went from `private` to `const val` with a KDoc saying why it's public, so there's one copy of the distinction rather than two.

I verified both of your paths against the code before changing anything, and they hold:

- `parseSinglePlaylistFromTwoRowRenderer` accepts any `VL`-prefixed browseId except `VLLM`/`VLSE` and does `browseId.removePrefix("VL")` (`YTMusicApiClient.kt:1024-1027`), and `getPlaylistTracks` already documents that "getUserPlaylists() can also return radios that the user has saved" (`:212-215`).
- `homeFeedMixIds` is written at exactly one site, inside the Success branch (`PlaylistFetchWorker.kt:455`), so `Empty`/`Error`/catch/discovery-off all leave it empty.

I'd also talked myself out of exactly this guard earlier on "the rule is one-way so the row settles" grounds. It does settle — your point is that it settles *wrong*, and the one-way rule is what makes that permanent rather than transient. That was the part I had backwards.

## On the test ask

You were right that the truth table couldn't catch this — the function was correct for the pairs it was handed. Two changes:

1. `ReconciledPlaylistTypeTest` now covers source and id namespace: a Spotify-generated mix, Discover Weekly's separate sub-namespace, a saved YouTube radio, and a loop asserting no YouTube id can be re-typed.
2. `DiffWorkerTest` gets the three cases you described, driving real diffs against the in-memory DB so the guard is exercised where the arguments are *assembled*: the #437 heal (asserts `CUSTOM` and `mix_number == null`), a `37i9dQZEVX…` Discover Weekly row, and a `RDCLAK5uy_…` supermix row — the last two asserting both `type` and `mix_number` survive.

I checked these actually bite rather than just passing: with the guard neutered back to the two-enum form, all six negative cases fail and the heal case still passes.

```
DiffWorkerTest > diff does not re-type a saved youtube radio arriving as CUSTOM FAILED
DiffWorkerTest > diff does not re-type a spotify-generated mix arriving as CUSTOM FAILED
ReconciledPlaylistTypeTest > discover weekly is never re-typed FAILED
ReconciledPlaylistTypeTest > a saved youtube radio is never re-typed FAILED
ReconciledPlaylistTypeTest > no youtube id can be re-typed FAILED
ReconciledPlaylistTypeTest > a spotify-generated mix is never re-typed FAILED
```

Guard restored: green, and `:app:compileDebugKotlin` green.

## PR description

Corrected — defect (a) is `48e156da` (#415) on master and was never in this diff. Sorry for the noise; I'd read the working tree at an older commit than the branch tip and re-derived a fix that had already shipped.

I went with the namespace guard rather than plumbing a "home-feed pass succeeded" flag into `DiffWorker`, since it also covers the discovery-off case without a new input key. Happy to switch to the run-scoped version if you'd prefer the reconcile to stay conditional on the pass.

chaudhary-lakshay:
@rawnaldclark gentle nudge on this one when you have a moment — the requested change landed in `0675bf53` (branch head) back on the 24th, along with the test coverage you asked for. Branch is still `MERGEABLE` against master.

Recapping what changed so you don't have to re-read the thread:

- `reconciledPlaylistType` now takes `source` and `sourceId`, and only fires for Spotify ids outside the `SPOTIFY_OWNED_ID_PREFIX` namespace — so neither of the paths you traced (saved `VLRD…` YouTube auto-mixes, and Spotify mixes that don't match `Daily Mix \d+` when `homeFeedMixIds` is empty) can re-type a real mix.
- `ReconciledPlaylistTypeTest` covers the namespace, not just the enum pair; `DiffWorkerTest` exercises the guard where the arguments are assembled, against the in-memory DB.
- Verified the tests fail with the guard neutered back to the two-enum form (6 red), pass with it restored.

No rush if you're mid-something — flagging mainly because #452 is stacked on this branch, so it's blocked behind whatever you decide here. Happy to switch to the run-scoped variant instead if that's the shape you'd prefer.

rawnaldclark (approved):
Re-reviewed the branch head (`0675bf53`): `reconciledPlaylistType` now takes `source` + `sourceId` and only fires for Spotify ids outside `SPOTIFY_OWNED_ID_PREFIX`, `DAILY_MIX → CUSTOM`, one-way — exactly the guard I asked for, with the namespace covered in `ReconciledPlaylistTypeTest`. Sorry for the ten-day wait. Merging with a merge commit rather than a squash, since #452, #453 and #454 all contain these commits and a squash would make each of them re-show this diff.

PR #448, opened 2026-08-23
fix(sync): make the Synced/Off chips filter every synced row (#373)

### LINKED ISSUE #373: [BUG] The search filter in the Spotify playlist manager menu is not working.
### Before you start

### Stash version

0.9.85

### Phone & Android version

Samsung a16 5g android 16

### Where does it happen?

#### Summary
When selecting the **"Off"** filter option under the **Spotify playlists** management screen, active/enabled playlists (toggled ON) are still visible in the list instead of filtering exclusively for disabled/off playlists.

#### Expected Behavior
Only playlists that are currently toggled **OFF** (disabled) should be displayed in the list.

#### Actual Behavior
Both active (toggled ON) and disabled (toggled OFF) playlists remain visible regardless of selecting the **`Off`** filter.

### Steps to reproduce

1. Open the app and go to the **Sync tab > Spotify playlists** menu.
2. Note the current filter status (the default is `All`).
3. Tap the **`Off`** filter button.
4. Check the list of playlists displayed.

Image

chaudhary-lakshay:
Fixes #373.

## Root cause

`ManagePlaylistsScreen` applied the segment chip to the "Your playlists" list only:

```kotlin
val bySegment = when (segment) { ... customAll.filter { ... } }
val visibleLiked = liked?.takeIf { matchesQuery(it) }     // no segment
val visibleMixes = mixes.filter { matchesQuery(it) }      // no segment
val visibleCustom = bySegment.filter { matchesQuery(it) }
```

The Liked row and every mix row got the text query but never the segment, so tapping **Off** left rows with an ON switch on screen and the filter read as broken — which is what the reporter described ("both active and disabled playlists remain visible").

## Fix

The chips name a sync state, so they now reach every row whose switch **is** sync state — Liked included — through one shared predicate rather than an inline `when` duplicated per call site.

Mix rows are **excluded** rather than filtered. A mix row's switch is `hideFromHome` ("shown on Home"), a different axis; filtering mixes by `syncEnabled` would land every one of them under "Off" still wearing an ON-looking switch, i.e. the same confusion one level down. The section's header and its discovery switch render regardless of segment, so the "reachable before the first sync pulls any mixes in" guarantee added by #335/#344 is preserved.

## Verification

`ManageSegmentFilterTest` — the Off/Synced/All contract, mix rows under All only, and a guard that no chip except All can keep both states.

Not device-verified: showing the difference needs an account with a Liked row and at least one mix, and my test account has neither. The change is a pure-function extraction plus three call sites, and `:app:compileDebugKotlin` is green.

### DIFF
--- a/feature/sync/src/main/kotlin/com/stash/feature/sync/ManagePlaylistsScreen.kt
+++ b/feature/sync/src/main/kotlin/com/stash/feature/sync/ManagePlaylistsScreen.kt
@@ -93,18 +93,20 @@ fun ManagePlaylistsScreen(
         if (rows.isNotEmpty()) listState.scrollToItem(0)
     }
 
-    val bySegment = when (segment) {
-        ManageSegment.ALL -> customAll
-        ManageSegment.SYNCED -> customAll.filter { it.syncEnabled }
-        ManageSegment.OFF -> customAll.filter { !it.syncEnabled }
-    }
+    val bySegment = customAll.filter { matchesSegment(segment, it.syncEnabled) }
     // The query filters every section, not just "Your playlists": accounts can
     // have 100+ auto mixes, and an unfiltered Liked/Mixes block pushes the
     // filtered custom results off-screen - making search look broken (#310).
+    //
+    // The SEGMENT reaches every section whose switch is sync state, for the same
+    // reason: it used to filter the custom list alone, so "Off" left the Liked
+    // row and every mix on screen wearing an ON switch and read as broken
+    // (#373). Mix rows are excluded rather than filtered — their switch is a
+    // different axis (see showMixRows).
     val q = query.trim()
     fun matchesQuery(row: ManageRow) = q.isBlank() || row.name.contains(q, ignoreCase = true)
-    val visibleLiked = liked?.takeIf { matchesQuery(it) }
-    val visibleMixes = mixes.filter { matchesQuery(it) }
+    val visibleLiked = liked?.takeIf { matchesQuery(it) && matchesSegment(segment, it.syncEnabled) }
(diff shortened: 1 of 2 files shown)


rawnaldclark (approved):
Nice catch, and the fix reads correctly — the chips name a sync state, so applying them to every synced row is the right call. Approving.

I did check the semantics I was most worried about: whether gating mix rows on the segment would drop every mix permanently into the "Off" bucket, given mixes are governed by `hide_from_home` rather than `sync_enabled`. It holds up in practice, and the "empty mixes section" state was already reachable before this PR via a non-matching text query, so this doesn't introduce it.

Two small things, neither blocking:

**1. The test pins the predicates, not the call sites.**

`ManageSegmentFilterTest` calls `matchesSegment` / `showMixRows` directly, and nothing references `visibleLiked` / `visibleMixes` / `visibleCustom`. So deleting `&& matchesSegment(...)` from any of the three call sites (lines 108-110) reintroduces #373 with all five tests still green — which is exactly the shape of the bug being fixed here. The module has junit/mockk/coroutines-test but no Compose test dependency, so nothing else would catch it either.

Not asking for a Compose UI test; even hoisting the three visible-list computations into one small internal function and asserting *that* would pin the behaviour rather than the helpers.

**2. Two comments now contradict the code.**

- Line 154: `// -- Segment filter (applies to the custom "Your playlists" list) --`
- Line 281: `/** Segment filter over the custom "Your playlists" list. */`

Both were true before this PR and are false after it, since line 108 applies the segment to Liked and 109 gates mixes on it. Two-line fix.

Neither blocks merge — happy for these to land as a follow-up if you'd rather.

PR #449, opened 2026-08-23
fix(update): tell the user what the update check found (#377)

### LINKED ISSUE #377: [Bug]: Check for update
### Before you start

### Stash version

V0.9.84 / v0 9.85 / v0.986

### Phone & Android version

Android 15

### Where does it happen?

Other / not sure

Hi and thanks for your job !
Just i found a little bug on my device since i have your app. (I have it since V0.9.84 )
The Check for update function don't find any update. When i press it, there is a message " checking for update" and nothing else...
I use your app on Poco F5 with hyper OS 3.

### Steps to reproduce

Go to settings and presse Check for update button

Thanks for all your job ;)

chaudhary-lakshay:
Fixes #377.

## Root cause

"Check for updates" enqueued `UpdateCheckWorker`, toasted "Checking for updates…", and then never spoke again. Every no-update exit returned a bare `Result.success()`, so there was nothing for the UI to observe — the check ran fine and stayed invisible. The three reporters were all **on the latest release**, which is exactly the branch that hits `Result.success()` with only a `Log.d`.

## Fix

Each terminal path names its outcome in `outputData`, and the About screen watches the one-shot unique work and reports what it concluded: on the latest version, an update with its tag, or a failure.

Two more bugs fall out of the same change:

- **"Already notified for this tag" suppressed the answer, not just the duplicate notification.** Someone who dismissed the notice once and asked again got nothing back, even though an update existed. It now suppresses only the notification.

- **A manual check runs unconstrained.** With `NetworkType.CONNECTED`, an offline tap left the work ENQUEUED forever — never started, never terminal — so a UI waiting on a finished `WorkInfo` would wait for good. Someone who asked gets a fast "couldn't check"; background checks keep the constraint, because deferring until there's a network is right when nobody is waiting. (Caught by review before this shipped.)

Also strips an uppercase `V` in `isNewerVersion`: only lowercase was removed, so `V1.0.0` parsed as `[0, 0, 0]` and any remote tag looked newer than it.

## Verification

`UpdateCheckWorkerTest` — outcome mapping, and `isNewerVersion` across numeric-vs-text ordering (`0.10.0` > `0.2.0`, `0.9.9` < `0.9.84`), prerelease suffixes, missing components, and both `v`/`V` prefixes. This function had no tests before.

On-device, Samsung Galaxy Tab Active3 (SM-T570, Android 13), app at **0.9.99** — the reporters' exact state:

- **Online:** tap → `UpdateCheckWorker: Already on latest (0.9.99), remote is v0.9.99` → toast **"You're on the latest version (0.9.99)"**. Before this change that path produced the log line and nothing else.
- **Offline** (Wi-Fi and data off): tap → `UpdateCheckWorker: Update check failed` → toast **"Couldn't check for updates — try again later"**, and the button returns to its enabled state rather than sticking on "Checking…".

### DIFF
--- a/core/data/src/main/kotlin/com/stash/core/data/sync/workers/UpdateCheckWorker.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/sync/workers/UpdateCheckWorker.kt
@@ -18,6 +18,7 @@ import androidx.work.OneTimeWorkRequestBuilder
 import androidx.work.PeriodicWorkRequestBuilder
 import androidx.work.WorkManager
 import androidx.work.WorkerParameters
+import androidx.work.workDataOf
 import com.stash.core.data.sync.SyncNotificationManager
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.jsonObject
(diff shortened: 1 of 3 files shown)


ParaliyzedEvo:
@chaudhary-lakshay ik you've opened up prs to fix issues, but there is a backlog which also imports issues from discord instead of working of issues through here to help like speed up the process of fixes so we dont overlap
https://github.com/users/ParaliyzedEvo/projects/3

rawnaldclark (approved):
This is a good fix for #377 — naming the outcome in `outputData` and having About report it is the right shape, and the "already notified" interaction is handled correctly (the dedupe still suppresses repeat *notifications* without suppressing the in-UI result). Approving.

I specifically checked the classic failure mode for this pattern — a finished `WorkInfo` from a *previous* check being replayed the moment the user opens About, showing a bogus "you're on the latest" without them pressing anything. That doesn't happen here; the `checking` gate gets it right.

Two small things, neither blocking:

**1. `checking` should be `rememberSaveable`.**

`SettingsAboutScreen.kt:145` uses a plain `remember`, and `MainActivity` declares no `android:configChanges` and no fixed orientation, so a rotation recreates the Activity and resets `checking` to false — after which `if (!checking) return@collect` (line 154) drops the terminal `WorkInfo`. The user rotates during the ~1s in-flight window and the check goes quiet again, which is the exact complaint #377 is about, just narrower.

It's a one-word change, and the recreated screen does leave an enabled button rather than a dead end, so it's genuinely minor — but `rememberSaveable` closes it.

**2. None of the new tests cover what the PR changes.**

`UpdateCheckWorkerTest` imports only `isNewerVersion` and `outcomeFor`; it never constructs the worker or calls `doWork()`. `outcomeFor` is a companion helper independent of `doWork`, so reverting every terminal path back to a bare `Result.success()` leaves all 12 tests green — i.e. the tests can't fail on the regression this PR exists to prevent.

The part I'd most want pinned is the new `giveUpOrRetry` split at `UpdateCheckWorker.kt:259` (manual → `Result.failure`, background → `Result.retry`). That's the one genuinely risky new decision and it currently ships unasserted; getting it backwards means either a silent manual check or a background check that retries forever.

Neither blocks merge.

PR #450, opened 2026-08-23
Fix YouTube-fallback streaming reliability (stale cache, duplicate resolves, gated fast-client URLs) + sync saved YT Music albums

### LINKED ISSUE #206: Songs get skipped mid playthrough
I was out on a drive today and noticed that songs would skip while streaming a playlist. They would play part of the way through and then skip to the next song.

---
_Imported from Discord: https://discord.com/channels/1508402965042561054/1536179956504338513_

### LINKED ISSUE #444: [Bug]: online songs stopped playing
### Before you start

### Stash version

0.9.99

### Phone & Android version

Galaxy 25+

### Where does it happen?

Streaming (Online mode)

Songs stopped playing in online mode after v0.95.  Previously downloaded songs still work.

### Steps to reproduce

Play any song from homepage or library Playlist 

_No response_

ParaliyzedEvo:
## Summary
* Track playback would silently skip past a track whenever its stream had no lossless source and needed the YouTube fallback, even with the fallback enabled. Root cause turned out to be three separate, compounding bugs in the streaming-fallback path rather than a single one:
  1. **Stale cache reused on retry.** `StreamErrorCascadeGuard`'s `RetrySameItem` verdict re-prepared the same track without evicting its `StreamUrlCache` entry, so a URL that 403'd before its own `expiresAtMs` (a session/token-bound YouTube URL dying early) was served identically on the "retry," guaranteeing a second failure and escalating straight to a skip.
  2. **Duplicate concurrent resolves for the same track.** `StreamSourceRegistry.resolve()` had no single-flight protection (unlike `PreviewUrlExtractor`'s per-videoId coalescing one layer down), so a foreground tap and the next-up prefetch watcher could both miss `StreamUrlCache` within the same window and independently spin up two full resolver chains for the same track — doubling latency and jiosaavn/yt-dlp load, and in the worst case leaving playback stuck waiting on a redundant chain.
  3. **Gated URLs from yt-dlp's fast client slipping through unverified.** The `android_vr` fast lane's `best`-format fallback (used because that client advertises no reliable audio-only itag) increasingly lands on itag 18, a combined-AV stream that's PO-token-gated and 403s on real playback. Unlike the InnerTube lane (protected by `AudioUrlTailProbe`), the yt-dlp lane trusted any non-null URL outright. A live tail-byte probe was tried and rejected — verifying a session-bound URL can itself consume its one allowed fetch, causing the probe to pass and the following real request to fail. Fixed instead by distrusting the *format id*: only itags 251/250/140 (genuine audio-only) are trusted from the fast client; anything else (itag 18) now falls through to the slower, reliably-working default multi-client yt-dlp path.
* Separately: saved YouTube Music **albums** (e.g. an album saved from YT Music that also appears in the user's library) were never fetched by sync at all — only playlists, Liked Songs, and home mixes were. Added a new sync leg for the Library → Albums tab, a new `PlaylistType.ALBUM`, and a corresponding section in `ManagePlaylistsScreen` so saved albums sync and appear alongside playlists.

## Changes
* `core/media/src/main/kotlin/com/stash/core/media/PlayerRepositoryImpl.kt`
  * `onPlayerError`'s `StreamErrorCascadeGuard.Verdict.RetrySameItem` branch now calls `streamUrlCache.invalidate(trackId)` before `retryCurrentItem()`, so a retry actually re-resolves instead of replaying the same dead URL.
* `core/media/src/main/kotlin/com/stash/core/media/streaming/StreamSourceRegistry.kt`
  * Added single-flight coalescing to `resolve()`, keyed by `(trackId, allowYouTube, allowYtDlp)`, mirroring `PreviewUrlExtractor`'s existing per-videoId pattern. Original resolver-chain logic moved unchanged into a new private `doResolve()`.
* `data/download/src/main/kotlin/com/stash/data/download/preview/AudioUrlTailProbe.kt`
  * No net change — an intermediate `opens()` byte-0 probe was added and then removed after device testing showed it insufficient (and a full tail-based probe unsafe for this use) for the yt-dlp path.
* `data/download/src/main/kotlin/com/stash/data/download/preview/PreviewUrlExtractor.kt`
  * `runYtDlp()` now returns a `YtDlpResult(url, formatId)` instead of a bare URL, parsing the format id out of the existing `--print "fmt=..."` diagnostic line.
  * `extractViaYtDlp()` now only trusts the fast (`android_vr`) client's result when `formatId` is in a new `AUDIO_ONLY_ITAGS` allowlist (`251`, `250`, `140`); anything else (observed: itag 18) falls through to the default multi-client yt-dlp path.
* `core/model/src/main/kotlin/com/stash/core/model/PlaylistType.kt` (or wherever `PlaylistType` lives)
  * Added `ALBUM` case.
* `data/ytmusic/src/main/kotlin/com/stash/data/ytmusic/model/YTMusicModels.kt`
  * Added `PagedAlbums`.
* `data/ytmusic/src/main/kotlin/com/stash/data/ytmusic/YTMusicApiClient.kt`
  * Added `BROWSE_LIBRARY_ALBUMS` ("FEmusic_liked_albums", unverified against a live response — flagged for follow-up if it comes back empty for a user with saved albums).
  * Added `getSavedAlbums()`, `parseSavedAlbums()`, `parseSavedAlbumsContinuationPage()` (continuation shape unverified), and `parseSingleAlbumFromTwoRowRenderer()`.
* `core/data/src/main/kotlin/com/stash/core/data/sync/workers/PlaylistFetchWorker.kt`
  * Added `fetchAndSnapshotSavedAlbums()` and `fetchAndSnapshotSavedAlbum()`, run alongside the existing user-playlists fetch leg. Each saved album is snapshotted as a `RemotePlaylistSnapshotEntity` with `playlistType = PlaylistType.ALBUM`, tracklist pulled via `getAlbum(browseId)` (album browseIds don't resolve through `getPlaylistTracks`).
* `feature/sync/src/main/kotlin/com/stash/feature/sync/ManagePlaylistsScreen.kt`
  * Added an "Albums" section (opt-in per-album sync toggle, same as "Your playlists") between Mixes and Your playlists.

## Test plan
* [x] Device-tested across ~15 tracks needing YouTube fallback (no lossless credentials configured) — majority play correctly; remaining failures traced through three iterations of diagnostics logs to the fixes above.
* [x] Confirmed via logs: stale-cache retry no longer replays an identical dead URL (`StreamUrlCache.invalidate` fires before retry).
* [x] Confirmed via logs: no duplicate `"chain for <id>"` entries for the same track within a single playback attempt after adding resolve-level coalescing.
* [x] Confirmed via logs: fast-client itag 18 hits now log `"falling to default client"` and fall through instead of reaching `onPlayerError`.

Closes https://github.com/ParaliyzedEvo/Stash/issues/206
Closes #444


rawnaldclark:
Three separate, correctly-diagnosed bugs in one sitting — the duplicate-resolve one especially, since it only shows up as "this felt slow" unless you go looking. Thank you.

Heads-up on how I've handled this, because it's not the usual flow: **the three streaming fixes are now on `fix/youtube-po-token-client-chain` (commit `603b7fa7`), with you as co-author.** I implemented them rather than sending this back and waiting, because two of the three needed a correction I'd have had to explain at this length anyway, and the branch is the fix for a week-long all-users playback outage that I don't want to hold. The album feature is untouched and still yours — see the end.

Here's what changed and why, so nothing is a surprise.

## 1. Cache eviction before retry — correct, one guard changed

Diagnosis is exactly right: `StreamUrlCache` only expires on *time*, so a URL that 403s before its own `expiresAtMs` gets replayed identically by `LazyResolvingDataSource`'s cache-before-resolve, and the "retry" was never a second attempt. The KDoc claiming otherwise was simply untrue.

One change — the guard is now `!= 0L` rather than `> 0L`:

```kotlin
?.getLong(EXTRA_TRACK_ID, 0L)?.takeIf { it != 0L }
```

Radio and search-surface tracks carry synthetic ids of `videoId.hashCode()`, which are **negative roughly half the time** — and those streamed rows are the ones most likely to hit this path in the first place. With `> 0L` the fix would skip eviction for a large share of its own target population. `maybeStampCurrentItemQuality` about 200 lines below has this exact scar: it used to use `<= 0L` and the badge silently stayed wrong for radio tracks until it was changed to a 0-sentinel.

I also narrowed the comment's scope: this only helps `stash-resolve://` placeholder items. Once the prefetch upgrades a slot, `refreshControllerMediaItem` bakes the real URL into the MediaItem URI and the cache is never consulted — YouTube slots then self-heal via `RefreshingDataSource`, qbdlx/arcod slots don't. That remaining gap is a separate change.

## 2. Single-flight — correct, but the timeout had to move inside

The device evidence (a second full chain 325 ms later for the same track) is specific and convincing, and keying on `trackId + flags` is right — a background-fill call deliberately accepts a narrower source set and must never be handed a foreground result.

The correction is load-bearing. As written, the shared `Deferred` runs on `resolveScope`, so `LazyResolvingDataSource`'s `withTimeout(RESOLVE_DEADLINE_MS)` bounds only the **caller's `await()`** — not the detached job. A wedged resolver chain would then run forever, `invokeOnCompletion` would never fire, the map entry would never clear, and **every later attempt at that track would coalesce onto the same dead `Deferred` and time out again.** A transient hang becomes a permanently unplayable track, which is a worse failure than the duplicate resolve. So the timeout now lives inside the coroutine:

```kotlin
resolveScope.async(start = CoroutineStart.LAZY) {
    withTimeout(LazyResolvingDataSource.RESOLVE_DEADLINE_MS) { doResolve(...) }
}
```

Also: the `putIfAbsent` loser is now cancelled. `async` attaches the child to the scope's job at *construction*, even with `CoroutineStart.LAZY`, so an unstarted-and-uncancelled loser lingers in the parent's child list with its captured `TrackEntity` for the process lifetime. (Same latent leak exists in `PreviewUrlExtractor`'s coalescing, where the pattern came from — worth a follow-up.)

Worth knowing: this makes `prefetchInFlight` in `PlayerRepositoryImpl` redundant — the registry-level guard is a strict superset, since it also covers tap-vs-prefetch. Good deletion for a follow-up PR.

## 3. The gated-URL fix — your idea kept, the mechanism changed

"Distrust the format" is right and it's now in, as a free pre-filter ahead of the probe. Two adjustments:

**Tested on `vcodec`, not an itag allowlist.** The selector ends `bestaudio/best`, so a legitimate audio-only answer can be 139, 249, 774, or whatever YouTube adds next — a static `{251,250,140}` list throws those away. `vcodec` is already printed on the same `--print` line and is correct for every itag:

```kotlin
if (stream.vcodec != null && stream.vcodec != "none") { /* next client */ }
```

**The tail probe stays, and runs after it.** This is where I have to disagree with the PR body, and I want to be precise rather than hand-wavy about it, because you tested something real.

Your text says a live probe is unsafe because a session-bound URL's one allowed fetch can be consumed by the verification request. I checked that against the tree and I don't think it holds:

- On `master` there is **no `--print size=`** anywhere in `PreviewUrlExtractor`, and the only `tailProbe.servesFullFile` call site is the InnerTube lane. The filesize print that makes a clen-based probe of a yt-dlp URL possible doesn't exist there — it's added by the branch. So a "clen-based tail probe accepted the URL" run wasn't possible on that base.
- Your own body says what was measured: *"an intermediate `opens()` byte-0 probe was added and then removed after device testing showed it insufficient."* A **byte-0** probe passing and playback 403ing later is the *expected* result — the gate is positional (`PreviewUrlExtractor` documents "every offset past ~1.1 MB 403s, via Range header, `&range=` query, and sequential chunking alike"), and `AudioUrlTailProbe`'s KDoc says exactly this: *"A HEAD or head-range request cannot tell a gated URL from a good one. The last byte can."* So the byte-0 result is evidence for the last-byte probe, not against it.
- googlevideo URLs don't appear to have a one-fetch property at all: `RefreshingDataSourceFactory` is built on re-opening the same URL at new offsets, and Media3 spawns a fresh DataSource per load/seek sharing one URL. If a single fetch burned it, every streamed source in the app would be broken, not just YouTube.
- And the PR keeps the InnerTube tail probe while calling that lane "protected by `AudioUrlTailProbe`" — a probe can't both burn the URL and protect the lane.

Empirically, on the branch: repeated cold resolves log `web_embedded served a full-file URL` and then play — one track observed with 96 s buffered, another to ~3:15 of 3:37, zero 403s across several restarts. If the probe consumed the fetch, playback would 403 immediately every time.

The two checks cover different things, which is why both are in: your format test catches a combined-AV answer with **no network call**, and the probe catches a genuinely audio-only URL that's nonetheless gated — the case the format test has no signal for, and the one that arrives the next time YouTube rotates the gate.

For context on why the chain rather than the allowlist alone: with `android_vr` still pinned, it now fails the allowlist on essentially every track under the PO-token wall, so every play would pay ~3.6 s for a doomed run and then ~12.8 s for the default client set, forever — and never regain audio-only serving. `web_embedded` serves itag 251 audio-only directly.

## The album feature — please split it into its own PR

It's ~315 of the 443 added lines, and none of it belongs in a change fixing a live outage; that change needs to stay independently revertable. Sending it back rather than adopting it, because it needs decisions that are yours to make. Blockers I'd want addressed:

- **Unbounded fan-out.** `paged.albums.map { async { fetchAndSnapshotSavedAlbum(...) } }.awaitAll()` takes no per-album permit, while the caller holds one for the whole leg — and the KDoc says the opposite ("gated by the same `sem` semaphore via the caller"). 40 saved albums fires 40 concurrent InnerTube `getAlbum` calls against `MAX_PARALLEL_YT_FETCHES = 3`. Move the permit inside, matching the mixes and user-playlists legs.
- **CI will go red.** `PlaylistTypeTest` asserts the exact enum set and is a deliberate migration tripwire; the PR adds `ALBUM` and ships no test changes. (The migration answer itself is fine — `type` is persisted by name into TEXT, no new column, no migration needed. Worth stating in the body.)
- **Pagination stops at page 1 and reports SUCCESS.** `parseSavedAlbums` walks a `gridRenderer`, but `extractContinuationToken` has no grid case — so a truncated inventory feeds the REFRESH-mode `deactivateMissingForSource` gate, which is destructive. Until the continuation shape is verified against a live paginated response, force `partial = true` whenever page 1 returns a full grid.
- **`ALBUM` has no display surface.** There's no exhaustive `when` over `PlaylistType` anywhere, so it produces zero compile errors and then silently misses `LibraryViewModel`'s `== CUSTOM` filter — albums become toggleable on Sync but invisible in Library, and the two screens' counters disagree. My suggestion: **don't add a `PlaylistType` at all** — snapshot saved albums as `CUSTOM`. The stated goal is that they sync; a new enum case buys nothing yet and costs a display surface, a counter reconciliation and a tripwire update.

Two things that are clean and worth saying: this is **not** a repeat of #447's re-typing problem (ALBUM rows are stably typed), and the #368 mass-download vector is closed because `defaultSyncEnabled` returns false and the new section ships no "Enable all". Both good instincts.

Also note the PR now conflicts with master regardless — #448 landed and touched `ManagePlaylistsScreen`, and the albums hunk uses pre-#448 context lines. The new Albums section would reintroduce the filtering bug #448 just fixed, so that section needs `matchesSegment` applied like the others.

Closing this one as superseded on the streaming side, or happy to leave it open if you'd rather rebase it down to just the albums work — your call. Either way the album feature is yours and I'd like to see it land properly.

ParaliyzedEvo:
@rawnaldclark rawn understood, making a new pr with your fixes in mind

PR #451, opened 2026-08-23
Sync saved YouTube Music albums

### LINKED ISSUE #113: Sync YT albums
As the title suggests, albums from YouTube do not get synced, I am not sure if this is the supposed behavior, if it is, is there a way for it to be integrated into the app? Thanks.

ParaliyzedEvo:
## Summary
Saved YouTube Music albums (e.g. an album saved from YT Music that also
shows up in the user's library) were never fetched by sync — only
playlists, Liked Songs, and home mixes were. This adds a sync leg for the
Library → Albums tab so saved albums sync and appear in Manage Playlists
alongside custom playlists.

Split out from #<streaming-fix-PR-number> per review — this is pure
feature work and shouldn't ride along with a live-outage fix.

Snapshots saved albums as `PlaylistType.CUSTOM` rather than introducing a
new enum case: a new `PlaylistType` needs a display surface everywhere the
app switches on the type (Library filtering, counters, etc.), and nothing
here needs that distinction yet. `CUSTOM` gets the album synced, toggle-able,
and diffed with no additional surface area.

## Changes
* `data/ytmusic/src/main/kotlin/com/stash/data/ytmusic/model/YTMusicModels.kt`
  * Added `PagedAlbums`.
* `data/ytmusic/src/main/kotlin/com/stash/data/ytmusic/YTMusicApiClient.kt`
  * Added `BROWSE_LIBRARY_ALBUMS` ("FEmusic_liked_albums") — **unverified
    against a live response**, see test plan.
  * Added `getSavedAlbums()`, `parseSavedAlbums()`,
    `parseSavedAlbumsContinuationPage()`, `parseSingleAlbumFromTwoRowRenderer()`.
  * `getSavedAlbums()` always returns `partial = true` until the grid
    continuation shape above is confirmed live — `extractContinuationToken`
    has no `gridContinuation` case yet, so a truncated fetch would otherwise
    report success and feed the destructive REFRESH-mode deactivation gate.
* `core/data/src/main/kotlin/com/stash/core/data/sync/workers/PlaylistFetchWorker.kt`
  * Added `fetchAndSnapshotSavedAlbums()` and `fetchAndSnapshotSavedAlbum()`,
    run alongside the existing user-playlists leg, gated on the shared
    `MAX_PARALLEL_YT_FETCHES` semaphore per-album (matching the mixes/
    user-playlists fan-out pattern).
  * Each saved album snapshots as a `RemotePlaylistSnapshotEntity` with
    `playlistType = PlaylistType.CUSTOM`; tracklist pulled via
    `getAlbum(browseId)` since album browseIds don't resolve through
    `getPlaylistTracks`.
* `feature/sync/src/main/kotlin/com/stash/feature/sync/ManagePlaylistsScreen.kt`
  * Added an "Albums" section (opt-in per-album sync toggle, same as "Your
    playlists") between Mixes and Your playlists, routed through the same
    segment-filtering fix from #448.

## Test plan
* [x] Confirm sync completes, the album appears in Manage Playlists under
      the new Albums section, and toggling sync on downloads/queues its
      tracks correctly.

Closes #113


rawnaldclark (changes requested):
The feature is wanted and the parser structure mirrors the playlist leg well. Three things stop it from merging as it stands, one of them a regression for every YouTube user.

## Blocking

1. **Every sync now disables refresh-mode cleanup for all YouTube playlists.** `getSavedAlbums()` hard-codes `partial = true`, and the worker turns that into `markYoutubeIncomplete(...)`. `youtubeInventoryComplete` is one flag for the whole YouTube leg, and `DiffWorker` gates deactivation on `shouldDeactivateMissingPlaylists(youtubeSyncMode, youtubeInventoryComplete)` — REFRESH *and* complete. So with this branch, a playlist the user deletes on YouTube Music is never hidden in Stash again, for everyone, on every sync, whether or not they have saved albums. That is the exact behaviour #343/#348 fixed. The album leg's uncertainty has to stay inside the album leg (e.g. skip the album snapshot when the grid paginates, or only mark incomplete when a continuation token was actually seen and not followed), never flip the shared flag.

2. **The type is inconsistent across the diff.** `Playlist.kt` adds `PlaylistType.ALBUM`; the worker snapshots albums as `PlaylistType.CUSTOM`; `ManagePlaylistsScreen` builds its new Albums section from `rows.filter { it.type == PlaylistType.ALBUM }`. The section can never populate, albums land under "Your playlists" with an `MPREb_` id in a slot every other code path assumes is a `VL` playlist id, and the PR body ("snapshots saved albums as CUSTOM rather than introducing a new enum case") describes the opposite of the code. Pick one: `ALBUM` everywhere (and handle it in the Library filter and counters), or `CUSTOM` and drop the enum case and the Manage section.

3. **The browse id is unverified.** Per your own note, `FEmusic_liked_albums` and the grid continuation shape have not been checked against a live response, so the feature is unproven end to end. One capture from a real account with more than one page of saved albums settles both the shape and item 1.

## Minor

- `albumsDeferred.await()` is indented as if it belongs to the `if (!likedDeferred.await())` line above it; it doesn't (the `if` has a single statement). Put it on its own line at the block's indentation so the next reader doesn't think it's conditional.

Once the live shape is confirmed and the flag stays local to the album leg, this is a straightforward re-review.

ParaliyzedEvo:
@rawnaldclark Everything should be fixed now and tested syncing with the yt albums. Nothing broke on my device and everything works as intended

rawnaldclark (commented):
Re-reviewed after the new commits. Two of the three blockers are resolved:

1. **Refresh-mode cleanup** — `partial` is now computed (`paginated.partial || gridTruncated`) and `markYoutubeIncomplete` fires only when the album leg was actually partial. That removes the regression for every YouTube user. Resolved.
2. **Type consistency** — albums are snapshotted as `PlaylistType.CUSTOM` throughout and the separate `ALBUM` type is gone. Resolved.

The one left is the live proof for `FEmusic_liked_albums` and the grid shape (`gridRenderer` → `musicTwoRowItemRenderer`, `MPREb_` browse ids). The browse id matches what ytmusicapi uses for the library albums page, so I expect it to be right, but the parser has never seen a real response in this PR. One of these unblocks the merge:

- a captured `browse` response for `FEmusic_liked_albums` from your account (titles can be redacted) committed as a test fixture with a parser test, or
- a screenshot of the Library after a sync on your device showing the saved albums arriving, plus the `getSavedAlbums` log line with the count.

Also still open from the first review: `albumsDeferred.await()` is indented as if it belonged to the `if` above it. One line.

With the proof in, this merges.

ParaliyzedEvo:
@rawnaldclark 
Confirmed live against my account (2 saved album, previously missing from the index):

Sync log: YOUTUBE · getSavedAlbums · SUCCESS · http=null · items=2
image — albums (top row) now appears in Library after this sync

This confirms FEmusic_liked_albums resolves correctly and parseSavedAlbums's grid-renderer shape (musicTwoRowItemRenderer, MPREb_ browse ids) parses a real response correctly. I don't have an account with enough saved albums to generate a genuine multi-page response, so gridHasUnhandledContinuation's continuation-detection path is implemented defensively (matches the same shape pattern the existing playlist-grid continuation handling uses) but not device-verified against a real second page — flagging that explicitly rather than claiming it's proven.

PR #452, opened 2026-08-25
feat(backup): choose what an Import Backup restores (#235)

### LINKED ISSUE #46: [feature]: Add options to the Import Backup feature
*(no text content)*

**Attachments:**
https://cdn.discordapp.com/attachments/1523300451393601666/1523300451557310616/Screenshot_20260705_161039.jpg?ex=6a557ed9&is=6a542d59&hm=2a0b3c2d2ddb6dadf8973647fc592a60cc5836eb9ed15e6cf830468362a6e59e&

---
_Imported from Discord: https://discord.com/channels/1508402965042561054/1523300451393601666_

### LINKED ISSUE #235: [feature]: Add options to the Import Backup feature
Currently when you import a backup, it will overwrite library and settings. Could there be a feature where you can choose what to import from the backup.

For e.g., if you want to import your library only, you select Import Library, and all your existing settings stays untouched.

Also, could it be that when you import the library, your existing playlists/songs doesn't get affected/removed, like importing a library from a backup only adds the songs from that backup.

For e.g., before importing a backup I have song 'X' in my library. In the backup there's song 'Y'. Currently, importing the backup will remove song 'X' and add song 'Y' only. I'm requesting for the Import option to only add song 'Y', making your final library have both songs. [that's the best I can explain it, ps. IGNORE IF THIS IS ALREADY THE CURRENT BEHAVIOR IN STASH]

Thanks.

chaudhary-lakshay:
Closes #235, Closes https://github.com/ParaliyzedEvo/Stash/issues/46.

Import Backup was all-or-nothing: it replaced the library **and** settings. This adds a scope picker so a backup can be merged instead of replacing, and so library and settings can be restored independently.

## Scopes

`BackupImportScope` covers both halves of what the issue asked for:

| Scope | Effect |
|---|---|
| `LIBRARY_MERGE` (default) | Adds the backup's missing songs/playlists/memberships. Nothing deleted, settings untouched, **no restart** — it writes through the live Room instance. |
| `LIBRARY_REPLACE` | Swaps in the backup's database, keeps current settings. |
| `SETTINGS_REPLACE` | Restores settings, keeps the current library. |
| `EVERYTHING_REPLACE` | The historical behavior. |

Merge identity is `spotifyUri → youtubeId → canonical(title, artist)`. Playlists match on `sourceId` and never inherit `syncEnabled` from a backup (keeps the #438 download-consent rule).

## Verified on device

Samsung Tab (SM-T570), debug build:

- Dialog shows 4 options, Merge preselected, the three destructive ones in red; the confirm button relabels `Merge` → `Overwrite All` and adds a "This cannot be undone." warning.
- **The issue's exact scenario:** library held song X, merged a backup containing song Y, ended with **both**. Tally read *"added 1 new track, 1 new playlist, and 1 playlist entry. Settings were left untouched."* and the process id was unchanged — no restart.
- Replace-everything left only the backup's track and did restart.

## Review fixes

Two defects found in review and fixed here, each with a regression test and a negative-control run:

- The membership merge guarded only the live side and never `ref.removedAt`, so a membership the user had **removed** in the backup was re-inserted as active — the mirror image of what merge promises.
- Merged tracks inherited `is_downloaded`/`file_path`/`file_size`. A backup ZIP carries no audio and merge skips the restart that runs the disk-truth sweep, so rows pointed playback and storage totals at files that aren't on this install. They now land not-downloaded; adoption re-claims them if the audio is really there.

## Tests

`core:data` 715 and `feature:settings` 38, 0 failures; `:app:compileDebugKotlin` clean.

### DIFF
--- a/core/data/src/main/kotlin/com/stash/core/data/db/DatabaseBackupManager.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/db/DatabaseBackupManager.kt
@@ -580,9 +580,24 @@ class DatabaseBackupManager @Inject constructor(
                         continue
                     }
 
-                    // id = 0 → fresh autoincrement PK; every other column —
-                    // including like timestamps and play stats — travels as-is.
-                    val newId = trackDao.insert(track.copy(id = 0))
+                    // id = 0 → fresh autoincrement PK. Like timestamps, play
+                    // stats and match metadata all travel as-is; the LOCAL
+                    // FILE columns must not. A backup ZIP carries no audio,
+                    // and merge deliberately skips the restart that would run
+                    // the disk-truth sweep, so an inherited is_downloaded
+                    // would leave rows pointing playback, storage totals and
+                    // the library-move flow at files that aren't there.
+                    // Landing as not-downloaded is the recoverable direction:
+                    // if the audio IS present, adoption re-claims it on the
+                    // next reconcile.
+                    val newId = trackDao.insert(
+                        track.copy(
+                            id = 0,
+                            isDownloaded = false,
+                            filePath = null,
+                            fileSizeBytes = 0,
+                        )
+                    )
                     track.spotifyUri?.let { liveBySpotifyUri.putIfAbsent(it, newId) }
                     track.youtubeId?.let { liveByYoutubeId.putIfAbsent(it, newId) }
                     if (hasCanonicalIdentity(track.canonicalTitle, track.canonicalArtist)) {
--- a/core/data/src/main/kotlin/com/stash/core/data/db/DatabaseBackupManager.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/db/DatabaseBackupManager.kt
@@ -2,13 +2,18 @@ package com.stash.core.data.db
 
 import android.content.Context
 import android.net.Uri
+import androidx.room.Room
+import androidx.room.RoomDatabase
+import androidx.room.withTransaction
 import androidx.sqlite.db.SimpleSQLiteQuery
+import com.stash.core.data.db.entity.PlaylistTrackCrossRef
 import dagger.hilt.android.qualifiers.ApplicationContext
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 import java.io.File
 import java.io.FileOutputStream
 import java.io.FileInputStream
+import java.time.Instant
 import java.util.zip.ZipEntry
 import java.util.zip.ZipInputStream
 import java.util.zip.ZipOutputStream
@@ -30,12 +35,97 @@ import kotlinx.serialization.json.Json
 private val SQLITE_MAGIC: ByteArray = "SQLite format 3".toByteArray(Charsets.US_ASCII) + 0
 
 /**
+ * What the user chose to restore from a backup ZIP (issue #235). The classic
+ * full overwrite is [EVERYTHING_REPLACE]; the three new scopes exist so users
+ * can take just their library or just their settings from a backup, and so a
+ * library restore can MERGE instead of replace.
+ *
+ * @property consumesLibraryDatabase  True when the import stages and reads
+ *   the backup's `stash.db` (either to swap it in or to copy rows out of it).
+ * @property swapsDatabaseFile        True when the live database FILE is
+ *   replaced — implies [consumesLibraryDatabase]. Merge never swaps; it
+ *   writes through the still-open live Room instance instead.
+ * @property restoresSettingsFiles    True when DataStore files are restored.
+ */
+enum class BackupImportScope(
+    val consumesLibraryDatabase: Boolean,
+    val swapsDatabaseFile: Boolean,
+    val restoresSettingsFiles: Boolean,
+) {
+    /**
+     * Adds the backup's missing songs/playlists/memberships INTO the current
+     * library. Nothing is deleted, nothing is overwritten, settings files are
+     * untouched, and — because the live Room instance stays open — NO restart
+     * is required.
+     */
+    LIBRARY_MERGE(
+        consumesLibraryDatabase = true,
+        swapsDatabaseFile = false,
+        restoresSettingsFiles = false,
+    ),
+
+    /** Swaps in the backup's database but keeps the current settings. */
+    LIBRARY_REPLACE(
+        consumesLibraryDatabase = true,
+        swapsDatabaseFile = true,
+        restoresSettingsFiles = false,
+    ),
+
+    /** Restores settings/preferences but keeps the current library. */
+    SETTINGS_REPLACE(
+        consumesLibraryDatabase = false,
+        swapsDatabaseFile = false,
+        restoresSettingsFiles = true,
+    ),
+
+    /** The historical behavior: replace both the database and the settings. */
+    EVERYTHING_REPLACE(
+        consumesLibraryDatabase = true,
+        swapsDatabaseFile = true,
+        restoresSettingsFiles = true,
+    ),
+}
+
+/**
+ * Outcome of a successful [DatabaseBackupManager.importDatabase].
+ *
+ * @property restoredTreeUri  External storage tree URI found in restored
+ *   preferences, when the scope restored settings files and one was present.
+ *   Null otherwise (including for library-only scopes — settings were never
+ *   touched, so there is nothing to re-grant).
+ * @property requiresRestart  True when files behind live singletons (Room /
+ *   DataStore) were swapped and the process must be killed to reload them.
+ *   False for [BackupImportScope.LIBRARY_MERGE], which writes through the
+ *   still-open live database.
+ * @property addedTracks      New track rows inserted (merge scope only).
+ * @property addedPlaylists   New playlist rows inserted (merge scope only).
+ * @property mergedMemberships Playlist memberships appended/reactivated
+ *   (merge scope only).
+ */
+data class BackupImportResult(
+    val restoredTreeUri: Uri?,
+    val requiresRestart: Boolean,
+    val addedTracks: Int = 0,
+    val addedPlaylists: Int = 0,
+    val mergedMemberships: Int = 0,
+)
+
+/** Internal tally of what one merge pass inserted. */
+private data class MergeCounts(
+    val addedTracks: Int,
+    val addedPlaylists: Int,
+    val mergedMemberships: Int,
+)
+
+/**
  * Handles exporting and importing the internal Room database, DataStore settings,
  * and encrypted tokens. Bundles everything into a single ZIP archive.
  *
  * Exporting uses a checkpoint-then-zip approach to ensure WAL-mode
- * consistency. Importing replaces the database and preferences files on disk
- * and requires an app restart to take effect.
+ * consistency. Importing replaces the database and/or preferences files on disk
+ * (per [BackupImportScope]) and normally requires an app restart to take
+ * effect — except the library-MERGE scope, which copies rows into the live
+ * database transactionally and needs no restart.
  */
 @Singleton
 class DatabaseBackupManager @Inject constructor(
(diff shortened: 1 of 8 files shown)


ParaliyzedEvo:
@chaudhary-lakshay don't forget to add `MIGRATION_42_43,` in `ALL_MIGRATIONS` in the `StashDatabase.kt` app will crash when trying to convert data and import it (@rawnaldclark still has to approve of this though, I'm not in charge of approving prs)

chaudhary-lakshay:
Good catch, thanks — you're right about the hazard, though I think it's a merge-ordering one rather than a bug in this branch as it stands.

Both `master` and this branch are at `version = 42`, highest migration `MIGRATION_41_42`, so `ALL_MIGRATIONS` is complete here today. `MIGRATION_42_43` comes from #439 (artist photos), which hasn't landed yet. So whichever of #452 / #439 merges second is the one that has to append it — and if it doesn't, the crash you're describing is exactly right: `DatabaseBackupManager` opens a throwaway `StashDatabase` over the staged backup to roll it forward before a merge-import, and a missing link in the chain takes the import down with "A migration from 42 to 43 was required but not found". The DI singleton hits the same wall on upgrade.

The array's KDoc already asked for the append in prose, which doesn't fail a build. So `19b214c2` adds `AllMigrationsChainTest`: it asserts `ALL_MIGRATIONS` is an unbroken chain whose last edge lands on the version Room actually reports for the built database. The version is read back from a real in-memory `StashDatabase` rather than restated in the test, so there's still exactly one place it's declared (`@Database` is `@Retention(CLASS)`, so the annotation isn't readable by reflection).

Verified it bites rather than just passing:

- drop `MIGRATION_41_42` from the array (your exact scenario) → FAILED, "ALL_MIGRATIONS stops short of the current schema version"
- drop `MIGRATION_20_21` → FAILED, "gap in the migration chain at version 20"
- restored → green

One thing I deliberately didn't do: add the literal `MIGRATION_42_43,` line. The val doesn't exist on this branch, so it wouldn't compile. The test is the durable version of your comment — whoever merges second gets a red build instead of a crash on someone's phone.

ParaliyzedEvo:
@chaudhary-lakshay ah got it thanks. my mistake since i was compiling beta builds for the discord server on my fork and added fixes from the prs here and mine. we also have a task board [here](https://github.com/users/ParaliyzedEvo/projects/3) if that can help with tracking issues

my fork: https://github.com/ParaliyzedEvo/Stash

chaudhary-lakshay:
Makes sense — that explains it. On a fork carrying both #452 and #439 the chain really is broken, so you were seeing a real crash, just one that needs both branches to exist.

Thanks for the board link, I'll keep an eye on it. And whichever of the two lands second now gets a red build instead of a broken import, so the fork case is covered too.

rawnaldclark:
Reviewed after #447 lands, since this branch contains its commits; the diff then shrinks to your own four. Two things already clear from a read: the scope model and the two review fixes (removed memberships, `is_downloaded` not inherited) are right, and `AllMigrationsChainTest` is a genuinely good addition — it turns the #439 ordering hazard Evo hit on his fork into a red build instead of a crashing import. Whichever of #452 / #439 merges second appends `MIGRATION_42_43`.

rawnaldclark:
Heads-up after today's master push: master now carries a schema migration to **version 43** (`MIGRATION_42_43`, data-only — rewrites stored YouTube `sddefault` thumbnail URLs to `hqdefault`; it landed with `2b0be6b5`, schema `43.json` committed).

This PR moves migration registration into `StashDatabase.ALL_MIGRATIONS` and adds `AllMigrationsChainTest`, so after a rebase onto master the list needs `MIGRATION_42_43` appended — otherwise the chain test goes red and, more importantly, `DatabaseModule` would stop registering it. One line. Everything else here can be reviewed as-is; I'll do the pass once the rebase is up.

rawnaldclark:
One more detail so the rebase doesn't fight the squash: this branch still carries #447's two commits (`380177c4`, `0675bf53`), and #447 went into master as a squash, so a plain `git rebase master` will replay them into conflicts. Drop them instead:

```
git fetch origin master
git rebase --onto origin/master 0675bf53
```

That replays only the four backup commits, and the only expected conflict is the `ALL_MIGRATIONS` list (add `MIGRATION_42_43` at the end).

rawnaldclark (commented):
Read the merge engine end to end against the entities and DAOs on master. The core is right: identity priority mirrors `findByAnyIdentity`, both unique columns are checked before any insert, playlists match on the unique `source_id` index, `insertCrossRef` is REPLACE so the soft-delete reactivation actually lands, the identity maps are built inside the same transaction that writes, and the throwaway Room instance is read to completion before the live transaction opens (no cross-instance lock). The six Robolectric cases cover the behaviors I would have asked for.

**Before merge**
1. Rebase per the comment above (`git rebase --onto origin/master 0675bf53`) and append `MIGRATION_42_43` to `ALL_MIGRATIONS`.
2. Declare the array as `val ALL_MIGRATIONS: Array<Migration> by lazy { arrayOf(…) }`. That deletes the "MUST stay the last member of this companion" rule: a lazy array captures every `MIGRATION_*` at first use, so declaration order stops mattering and the next contributor can't silently build it from an uninitialized val.

**One question, not blocking**
Every merged membership is written with `locallyAdded = true`, including memberships of Spotify/YouTube-synced playlists. For those, the remote list is the truth, and REFRESH exists to drop what the remote no longer has. Forcing the flag makes every restored membership permanently immune to that cleanup, so a stale six-month-old backup pins tracks into a synced playlist forever. Would `locallyAdded = ref.locallyAdded || <playlist is user-created / STASH_LIKED>` be closer to what a user restoring a library expects? Memberships the user added locally would still survive, and synced ones would follow the next sync as before.

**Nits**
- `SETTINGS_REPLACE` on a ZIP with no `datastore/` entries reports success and restores nothing. A "this backup contains no settings" failure would match the DB path's "contains no database" check.
- A merged track keeps `qualityKbps` / `sampleRateHz` from the backup while landing as not-downloaded, so quality badges can describe a file that isn't there. Resetting them alongside `filePath` keeps the row honest until adoption re-stamps it.
- A backup older than the canonical-identity columns gets those columns from the migration chain, not from the app-side backfill, so canonical matching may be blank for it and local-import tracks (no URI, no video id) can duplicate. Worth a line in the KDoc as a known limit.

chaudhary-lakshay:
Rebased onto master and worked through all four items. Not pushed yet — the SHAs below are local, so this lands as a force-push once I have the go-ahead.

**Rebase.** Your `--onto` recipe was right about the shape but #447 actually landed as a merge commit (`0357b653`), so both `380177c4` and `0675bf53` are already ancestors of master — `git rebase --onto origin/master 0675bf53` and a plain rebase are equivalent here. Used your form. One conflict, exactly where you predicted: the `ALL_MIGRATIONS` list.

**`MIGRATION_42_43`.** Appended, but derived rather than transcribed — master has moved several commits past `2b0be6b5` and I didn't want to assume 43 was still the tail. Master's `addMigrations` and this branch's `ALL_MIGRATIONS` are now an identical set, 40 entries each, checked by set-diff. Worth noting `AllMigrationsChainTest` would *not* have caught a dropped middle edge here: it asserts contiguity and the endpoint, and the endpoint was consistent by construction either way.

**`by lazy`.** Taken. The "MUST stay the last member" paragraph is gone.

**The `locallyAdded` question — you were right, and I got the fix wrong twice before it settled.**

My first answer was that your `|| <user-created / STASH_LIKED>` half was unnecessary, because `clearSyncedPlaylistTracks` is the only deletion consumer and never runs against a non-synced playlist. That's wrong. `SyncUndoDao.clearSyncedMembershipForRestore` also deletes `locally_added = 0`, across every playlist `capturePlaylists` snapshotted — and that's `SELECT :syncId, id, … FROM playlists`, no filter, no playlist-type test, no `removed_at` guard. So a `0` imported into a playlist no sync touches is deletable by "Undo last sync". Second leg: `MIGRATION_22_23` added the column `DEFAULT 0`, so every membership in a pre-v0.9.23 backup reads `0` — including the user's own hand-adds. Its migration comment reasons only about `clearPlaylistTracks`/REFRESH, which is stale w.r.t. `SyncUndoDao`.

The discriminator isn't playlist type, though. `type` doesn't separate them — imported Spotify playlists are `CUSTOM` (`PlaylistFetchWorker.kt:854`); nor does `source` or `syncEnabled`, since `createPlaylist` uses `BOTH`/`true`; nor a `custom_` prefix on `sourceId`, because `ensureCustomPlaylist` is called with ids like `"album:youtube:MPREb_xxx"` (`AlbumDiscoveryViewModel.kt:319`).

What does work is `last_synced`. `updateLastSynced` has exactly one call site, `DiffWorker.kt:876`, so non-null means a sync genuinely mirrored that playlist and its `0` rows are honest — REFRESH must still be able to drop what the remote dropped, which was your original point. Null means no sync has ever run, and there a `0` is a state no live add path can produce (they all stamp `1`), so it's read as user-added. Absence resolving to user-added is the safe direction: a wrongly-kept membership can be deleted by hand, a wrongly-dropped one is gone.

Two follow-on corrections on top of that:

- Ownership is read from `last_synced` on **either** side. A playlist synced in the backup but never yet here (sync enabled, not run) would otherwise have had its genuinely sync-written `0` rows forced to `1` — your original bug, in a narrower case. Decided in one pass over the backup→live id mapping so matched and newly-created playlists answer identically.
- The reactivation branch now ORs with the live row (`existing.locallyAdded || …`) instead of overwriting it. As written it let a backup's `0` downgrade a live hand-added row, which contradicts this function's own rule that existing rows keep their metadata.

**Nits.**

- *Quality columns* — `quality_kbps`, `sample_rate_hz` and `bits_per_sample` cleared alongside `file_path`. Also `album_art_path`: it's `file_path`'s sibling local path, not a cache key (`BlocklistGuard.kt:89` hands it to `fileDeleter.delete`), so a merged row was pointing at another install's files dir and would render no artwork despite a live `album_art_url` on the same row.
- *`SETTINGS_REPLACE` with no settings* — now fails. Deliberately keyed on `scope == SETTINGS_REPLACE` rather than `scope.restoresSettingsFiles`: the broader form also covers `EVERYTHING_REPLACE`, and the exporter emits `datastore/` entries only `if (datastoreDir.exists())`, so an export taken before any preference write would have failed a full restore whose database was fine.
- *Canonical-identity limit* — recorded in `mergeLibraryFrom`'s KDoc.

**Four things I found but did not fix**, since they're outside what you raised — your call whether they're this PR or follow-ups:

1. `metadata_embedded_at` survives the merge while `file_path` is nulled. Dormant now, but it bites after adoption: `getTracksNeedingEmbed` skips the row forever, so the newly adopted file never gets the tag set. One-word fix, no sweep to heal it.
2. `lyrics_fetched_at` asserts a `lyrics` row exists, and the merge copies only tracks, playlists and cross-refs — the `lyrics` table is never copied, so the `lyrics_fetched_at IS NULL` backfill skips it permanently.
3. `file_format` is stale on a merged row too, but **don't clear it blindly** — `getMissingQualityInfo` (`TrackDao.kt:875-885`) uses it as the recovery key for the very columns this PR nulls, recovering them only for rows whose format is in the lossless set.
4. `EVERYTHING_REPLACE` has no import coverage in `DatabaseBackupMergeTest` at all, and it's the default scope of the public API.

**Verification.** `DatabaseBackupMergeTest` 12/12, full `:core:data:testDebugUnitTest` green, `:app:assembleDebug` green. Each new test written and watched fail first — the membership ones fail on the exact provenance flip, not just on a count.

chaudhary-lakshay:
Pushed (force, since it's rebased) — `19b214c2...90994856`. The SHAs in my previous comment are now the real ones on the branch.

chaudhary-lakshay:
Shipped item 1 from my "found but did not fix" list — `metadata_embedded_at` is now cleared with the other file-derived columns.

Worth stating why it wasn't just untidiness: `getTracksNeedingEmbed` and `observeTracksNeedingEmbedCount` both gate on `metadata_embedded_at IS NULL`, so a stamp carried in from the backup is dormant while the row is not-downloaded and then permanently excludes it from the v0.9.35 tag pass once adoption re-claims the audio. Unlike the loudness columns there's no stale-stamp sweep to heal it.

The `copy` beside `filePath` now reads: `isDownloaded`, `filePath`, `albumArtPath`, `metadataEmbeddedAt`, `fileSizeBytes`, `qualityKbps`, `sampleRateHz`, `bitsPerSample`. The test that covers them is renamed to match the category rather than just quality.

Items 2-4 from that list I've deliberately left alone: `lyrics_fetched_at` (needs a call on whether the merge should copy the `lyrics` table at all, which is a bigger question than a null-out), `file_format` (clearing it breaks `getMissingQualityInfo`'s recovery of the columns above), and `EVERYTHING_REPLACE`'s missing import coverage. Say the word on any of them.

`DatabaseBackupMergeTest` 12/12, full `:core:data` green, `:app:assembleDebug` green, pushed.

PR #453, opened 2026-08-25
feat(downloads): selectable folder layouts + reorganize pass (#198, #104)

### LINKED ISSUE #104: one folder

### LINKED ISSUE #198: Feature: Define folder structure for library
Add ability to store songs in per playlist folder

chaudhary-lakshay:
Closes #198. Closes #104.

Downloads were always filed as `<Artist>/<Album>/<title>`. This makes the structure selectable and adds a one-shot pass to move already-downloaded tracks into the chosen shape.

## Layouts

| Layout | Shape |
|---|---|
| `ARTIST_ALBUM` (default) | `<artist>/<album>/<title>.<ext>` — unchanged historical behavior |
| `SINGLE_FOLDER` | `<artist>-<title>.<ext>` directly in the root (**#104**) |
| `PLAYLIST` | `<playlist>/<artist>-<title>.<ext>` (**#198**) |

`LibraryLayoutResolver` is the single source of truth for both writes (`resolve`) and lookups (`lookupCandidates`) — the layout was previously re-derived by hand in four places. The flat layouts prefix the artist because the directory no longer separates artists; two songs both called "Intro" would otherwise overwrite each other. Blank album → `singles/`; `PLAYLIST` with no sync-enabled owner falls back to Artist/Album.

Changing the setting never moves files. "Reorganize existing tracks" does that explicitly, and only appears when something is actually misplaced.

## Verified on device

Samsung Tab (SM-T570), debug build, internal storage:

- All three layouts listed, Artist/Album default, selection **persists across a full app restart**.
- With layout set to Single folder, a real download landed flat as `daft-punk-one-more-time-radio-edit.opus` with its `.lrc` beside it — new downloads follow the layout.
- Reorganize section stayed **hidden** while the file matched the layout and **appeared** the moment the layout changed and made it misplaced.

Not exercised on device: the reorganize pass itself against a real SAF/SD-card tree.

## Review fixes

Four defects found in review, all fixed here:

- **Wrong-song adoption.** The tree-wide filename fallback matched a bare `<title>.<ext>`. Outside its own `<artist>/<album>` directory that isn't an identity — a different artist's same-titled song could be adopted and this track's `file_path` healed onto the wrong audio. Only artist-qualified names match tree-wide now. Covered by two new tests.
- **SAF reorganize was broken entirely.** `moveWithinSaf` resolved the source with `DocumentFile.fromTreeUri`, but `file_path` holds the audio **document** URI — `fromTreeUri` rebuilds the tree root, so copy+delete ran against the wrong document for every SAF track. Uses `fromSingleUri`.
- **Cancel didn't cancel.** The per-entry `runCatching` swallowed `CancellationException`, so the pass kept moving files and then overwrote Idle with Done.
- **Disabled playlists claimed tracks.** `getFirstPlaylistNameForTrack` only *ordered* by `sync_enabled`, so a track whose memberships were all disabled was filed under a switched-off playlist. It now filters. Deliberately not the same rule as `DownloadQueueDao`'s "from &lt;playlist&gt;" label — showing *some* name is right for a label and wrong for a directory, and the KDoc no longer claims parity.

Lyrics sidecars also now follow the audio through a reorganize on **both** internal and SAF paths, and the SAF writer places the `.lrc` beside the audio's actual directory rather than re-deriving it from the current layout preference.

## Known gap

`ReorganizeLibraryCoordinator` has no dedicated unit test. Its resolver and matcher are covered, and the internal path was exercised on device, but the coordinator's own move logic is not under test — worth adding before this is relied on for large libraries.

## Tests

`core:data` 713, `feature:settings` 38, `data:lyrics` 40, `feature:sync` 45, `data:download` 36 — 0 failures; `:app:compileDebugKotlin` clean.

### DIFF
--- a/core/data/src/main/kotlin/com/stash/core/data/db/dao/TrackDao.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/db/dao/TrackDao.kt
@@ -332,6 +332,22 @@ interface TrackDao {
     @Query("UPDATE tracks SET file_path = :filePath WHERE id = :trackId")
     suspend fun healFilePath(trackId: Long, filePath: String)
 
+    /**
+     * Name of the playlist a track is filed under for the PLAYLIST library
+     * layout (issue #198): the first sync-enabled playlist containing the
+     * track, lowest playlist row id as tie-break — the exact rule the
+     * Failed Downloads viewer's scalar subquery uses so both surfaces agree
+     * on "which playlist owns this track". Null when the track belongs to
+     * no playlist; callers fall back to the Artist/Album location.
+     */
+    @Query("""
+        SELECT p.name FROM playlists p
+        INNER JOIN playlist_tracks pt ON pt.playlist_id = p.id
+        WHERE pt.track_id = :trackId AND pt.removed_at IS NULL
+        ORDER BY p.sync_enabled DESC, p.id ASC LIMIT 1
+    """)
+    suspend fun getFirstPlaylistNameForTrack(trackId: Long): String?
+
     /** Reverts tracks whose file no longer exists on disk back to
     *  "needs download" — clears is_downloaded, file_path, and the stale
     *  size so LibrarySizeHolder's next walk doesn't count phantom bytes. */
--- a/core/data/src/main/kotlin/com/stash/core/data/db/dao/TrackDao.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/db/dao/TrackDao.kt
@@ -334,17 +334,25 @@ interface TrackDao {
 
     /**
      * Name of the playlist a track is filed under for the PLAYLIST library
-     * layout (issue #198): the first sync-enabled playlist containing the
-     * track, lowest playlist row id as tie-break — the exact rule the
-     * Failed Downloads viewer's scalar subquery uses so both surfaces agree
-     * on "which playlist owns this track". Null when the track belongs to
-     * no playlist; callers fall back to the Artist/Album location.
+     * layout (issue #198): the first SYNC-ENABLED playlist containing the
+     * track, lowest playlist row id as tie-break. Null when no sync-enabled
+     * playlist owns it — callers fall back to the Artist/Album location.
+     *
+     * Deliberately NOT the same rule as [DownloadQueueDao]'s "from
+     * <playlist>" subquery, which merely ORDERs by `sync_enabled` so it can
+     * still show *some* name for a track whose only memberships are
+     * disabled. That is right for a label and wrong for a directory: filing
+     * audio under a playlist the user has switched off contradicts the
+     * documented fallback and puts files where they are not expected. A
+     * display name and a storage location are different questions, so this
+     * FILTERS where the viewer only sorts.
      */
     @Query("""
         SELECT p.name FROM playlists p
         INNER JOIN playlist_tracks pt ON pt.playlist_id = p.id
         WHERE pt.track_id = :trackId AND pt.removed_at IS NULL
-        ORDER BY p.sync_enabled DESC, p.id ASC LIMIT 1
+          AND p.sync_enabled = 1
+        ORDER BY p.id ASC LIMIT 1
     """)
     suspend fun getFirstPlaylistNameForTrack(trackId: Long): String?
 
--- a/core/data/src/main/kotlin/com/stash/core/data/db/dao/TrackDao.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/db/dao/TrackDao.kt
@@ -346,6 +346,12 @@ interface TrackDao {
      * documented fallback and puts files where they are not expected. A
      * display name and a storage location are different questions, so this
      * FILTERS where the viewer only sorts.
+     *
+     * A track in several enabled playlists is filed under the LOWEST
+     * playlist id, i.e. the one it was added to first. Ids never change,
+     * so the answer is stable: do NOT "fix" this to alphabetical name
+     * order — that would reshuffle files on disk every time a playlist is
+     * renamed.
      */
     @Query("""
         SELECT p.name FROM playlists p
--- a/core/data/src/main/kotlin/com/stash/core/data/library/FileExistenceChecker.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/library/FileExistenceChecker.kt
@@ -11,6 +11,9 @@ package com.stash.core.data.library
  * is present — the real implementation re-derives the file's location
  * from the current SAF tree grant using the same slug convention used
  * to write it, instead of trusting the old URI. See FileOrganizer.
+ * [FileExistenceChecker.exists]'s trackId lets the Per-playlist library
+ * layout (#198) re-derive the owning playlist exactly like the download
+ * path did.
  *
  * suspend because the SAF-tree lookup needs the current tree URI from
  * DataStore.
@@ -40,6 +43,7 @@ fun interface FileAdopter {
 
 fun interface FileExistenceChecker {
     suspend fun exists(
+        trackId: Long,
         artist: String,
         album: String?,
         title: String,
--- a/core/data/src/main/kotlin/com/stash/core/data/library/LibraryReconciliationUseCase.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/library/LibraryReconciliationUseCase.kt
@@ -68,8 +68,8 @@ class LibraryReconciliationUseCase @Inject constructor(
     suspend fun reconcile(
         syncId: Long? = null,
         onProgress: (step: Int, total: Int) -> Unit = { _, _ -> },
-        checkFileExists: suspend (artist: String, album: String?, title: String, filePath: String) -> com.stash.core.data.library.FileExistenceResult =
-            { _, _, _, _ -> com.stash.core.data.library.FileExistenceResult(exists = true) },
+        checkFileExists: suspend (trackId: Long, artist: String, album: String?, title: String, filePath: String) -> com.stash.core.data.library.FileExistenceResult =
+            { _, _, _, _, _ -> com.stash.core.data.library.FileExistenceResult(exists = true) },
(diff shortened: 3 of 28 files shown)


rawnaldclark:
Reviewed after #447 lands (same stacking). The four review fixes read correctly, and `LibraryLayoutResolver` as the single source of truth is the right shape. One gate before merge: this moves the user's files, and by your own note the reorganize pass was not exercised on a real SAF / SD-card tree and `ReorganizeLibraryCoordinator` has no unit test. I'll run it on a test device with an SD-card library before it goes in; a coordinator test that covers cancel-mid-pass and a failed move would make that a lot cheaper to trust.

rawnaldclark (commented):
Read `ReorganizeLibraryCoordinator`, `LibraryLayoutResolver`, the `FileOrganizer` changes and the reconciliation hook. The shape is good: the layout is a pure resolver with lookup candidates for all three layouts, so switching layouts without reorganizing doesn't make the disk-truth sweep mark tracks missing, and the DB-before-delete ordering is kept on both the internal and SAF paths. The `fromSingleUri` fix and the cancellation re-throw are both correct.

**Blocking: two rows can resolve to one target, and the pass deletes the first to make room for the second.**
`moveWithinInternal` does `if (target.exists()) target.delete()` and `moveWithinSaf` does `existing?.delete()` whenever the existing document isn't the source itself. The plan never checks that targets are distinct. Two library rows with the same artist and title slugs are common (studio and deluxe/live versions, a Spotify match and a YouTube match of the same song), and `SINGLE_FOLDER` drops the album segment, so every one of them collides there. Sequence for tracks A and B that both resolve to `artist-title.m4a`:

1. A is moved to the target, its row healed to that path.
2. B's move finds the target, deletes A's audio, copies B in, heals B to the same path.

Result: A's row now points at B's audio, and one file is gone. The download path has the same overwrite semantics today, but only for exact artist/album/title triples; the reorganize pass turns it into a library-wide sweep. Fix in `buildPlan`: group entries by `(dirKey, targetName)`, keep the first claimant (prefer the one already in place), and mark the rest as a "name clash" that is skipped and counted, before any delete runs. `LibraryLayoutResolverTest` should get the two-rows-one-target case.

**Suggestions**
- Refuse to start (or pause) while a library move or a FLAC swap is active. `SwapCoordinator` replaces a downloaded track's file and heals its path; if the reorganize moves the same track in the same window, whichever `healFilePath` runs last wins and the other file is orphaned.
- The PLAYLIST layout files a multi-playlist track under the lowest playlist id. That's deterministic, but adding a playlist with a lower id can't happen, so this is stable. Worth saying in the KDoc so nobody "fixes" it to name order, which would reshuffle files on every rename.

**Before merge**
This moves every file in a library, so it needs a run on a real SD/SAF tree, not only the internal path the unit tests cover. Once the collision guard is in, post the reorganize log from a device with a SAF tree (a handful of tracks in two playlists, at least one shared) and I'll run the internal-storage pass on the test rig here.

chaudhary-lakshay:
Blocking finding fixed, plus the coordinator tests you asked for. Not pushed yet, and this branch isn't rebased either — see the bottom — so I'm describing changes rather than pointing at SHAs you can't fetch.

**The collision.** Confirmed and fixed. `buildPlan` now groups entries by target and hands the name to one claimant — a row already sitting on it wins, otherwise plan order — and the rest are marked as a name clash, skipped and counted, decided in the plan so no delete ever runs for a loser. `countMisplacedTracks()` excludes them too, so the button doesn't promise moves that won't happen.

Written test-first, and it fails on exactly the sequence you described: `expected: STUDIO / but was: LIVE` on the target file, i.e. row A's audio replaced by B's with A's row healed to it.

One deviation from your instruction: you asked for the two-rows-one-target case in `LibraryLayoutResolverTest`, but the guard has to live in `buildPlan` (the resolver is per-track and can't see the batch), so the case is in a new `ReorganizeLibraryCoordinatorTest` instead. If you'd rather the resolver grew a batch-level API I'll move it, but that felt like more surface than the bug needs.

**One correction to the guard, from a second review pass.** My first version keyed the clash on `(dirKey, targetName)` alone, which treats an internal track and a SAF-tree track resolving to the same layout-relative name as a collision. They land in different storage roots and cannot overwrite each other, so the loser was being skipped forever and hidden from the misplaced count. The key is now scoped per storage root — verified by neutering the prefix, which reddens exactly that case and nothing else.

**Concurrent passes.** Your suggestion, and my first attempt at it was half-built: I had the reorganize check `MoveLibraryCoordinator`'s state, which only covers "move already running, reorganize starts". `MoveLibraryCoordinator.start()` never looked back, so starting the move *during* a reorganize left the identical race. A coordinator watching the other one can only ever cover one order, and covering both that way needs a dependency cycle.

So it's a shared `LibraryRewriteGate` instead: one holder at a time, the loser refused rather than queued (both passes are manual and restartable, so a queue would be a surprise), released in the launch's `finally` so cancel and failure hand it back. Symmetric by construction.

**The FLAC swap half I did not do.** `SwapCoordinator` exposes no state, so guarding against it means giving it one — new production surface rather than reading something that exists. The window is also much narrower than the move/reorganize pair, which are both whole-library sweeps. Happy to add it if you want it; it'd be `SwapCoordinator` claiming the same gate around the file replace.

**PLAYLIST lowest-id.** Recorded in `getFirstPlaylistNameForTrack`'s KDoc, including the "do not 'fix' this to name order" warning — that would reshuffle files on every playlist rename.

**Coordinator tests.** `ReorganizeLibraryCoordinatorTest` covers the clash, the in-place claimant winning, the storage-root partition, a failed move (counted, doesn't stop the pass), cancel mid-pass ending at Idle instead of overwriting it with a Done, and the gate in both directions plus its release on completion and on cancel. `MoveLibraryCoordinatorTest` covers the move refusing while the gate is held. 28 tests green across the four relevant classes, plus `:app:assembleDebug` for the new constructor arg.

**The device run is still yours.** Everything above is the internal path; the SAF branch needs a real provider. I can't produce the reorganize log from an SD-card tree you asked for, so that gate stands as you set it.

**Heads-up on this branch's CI, if you run it before the rebase.** `ArcodClientTest` has two failures here (`streamUrl throws on 429`, `streamUrl parses flat json url`), and the `:data:download` test task then hangs rather than finishing — it stalls after reporting them even when scoped to just that class. Both are pre-existing on this branch's stale base, not from anything in this PR: they reproduce on the base without my commits, and `origin/master` passes the same class clean (19 cases, ~4 min). Master's lossless work fixed it, so the rebase should take both away. Flagging it so a red run doesn't get read as this PR's.

**Rebase.** Blocked on my side by a local tooling permission, not by conflicts — I have your `--onto` recipe and it's the next thing to land.

chaudhary-lakshay:
Rebased and pushed. One conflict, in `TrackDownloaderImplDeferredTest` — both sides assert the same `deferIfInProgress(42L)`, only the comment wording diverged, so I took master's.

The CI heads-up in my previous comment is now moot, and confirmed as diagnosed: the full `:data:download:testDebugUnitTest` suite completes on the rebased branch — 83 classes, `ArcodClientTest` 19/19 — so both the two failures and the task hang were stale-base fallout, gone with master's lossless work. `:app:assembleDebug` green too.

Still yours: the SAF/SD-card device run.

rawnaldclark:
Re-read after the rebase. The collision guard is what I asked for: targets grouped in `buildPlan`, one claimant per name (the row already in place wins, then plan order), losers marked and skipped before any delete can run, and `countMisplacedTracks()` no longer promising moves that won't happen. The `LibraryRewriteGate` closes the move/swap window as well. Tests read right.

What's left is mine: the SAF/SD-card device run. It doesn't fit tonight's release (v0.9.104 tags in a few hours, and this pass moves every file in a library, so it gets a run on the test rig first, not the daily phone). It's first on the list for the release after, and I'll post the reorganize log here when it's done. Nothing further needed from you.

rawnaldclark:
Device run on a Pixel 6 (debug build of 9434418 merged onto master, 0 conflicts), 861 downloaded tracks on internal storage (857 FLAC, 4 Opus, 771 `.lrc` sidecars).

**Artist / Album → Single folder:** "Moved 861 tracks" in 5.7 s. All 861 rows point at flat `artist-title.ext` paths, 0 of them missing on disk, file count unchanged (1632). Sidecars travelled with their audio (every sampled FLAC still had its `.lrc` twin next to it).

**Playback of a moved file:** played a moved track from the Library tab; the app process holds the file open at its new flat path (`/proc/<pid>/fd`), so playback reads the moved file, not a stale copy.

**Single folder → Artist / Album:** "Moved 861 tracks" in 5.0 s. 851 paths came back byte-identical to the originals. The other 10 were downloaded before v0.9.77 with accent-stripped slugs (`bjrk/`, `ba/twilight/`) and were re-filed under the current slug (`björk/`, `bôa/twilight/`), which is the right outcome. The "Reorganize existing tracks" card disappears once everything is in place.

No `ReorganizeCoord` warnings in either pass. 19 nested `.lrc` files were left untouched; all 19 are pre-existing orphans (no audio beside them, no track row), so nothing to move.

**Not exercised:** the SAF (SD / USB) path; the test library is internal-only.

**Follow-up, not blocking:** emptied source directories are never pruned. After the Single-folder pass the tree still held 1,347 folders (747 empty, 600 holding only orphan sidecars or empty sub-folders), and after the round trip 1,360, with the old accent-stripped folders sitting empty next to their replacements. Invisible on internal storage, but an SD / USB library browsed over USB will show the husks, and that is this feature's headline case. A post-pass sweep that deletes emptied directories up to the music root (`File` internally, `DocumentFile` for SAF) would close it.

Merging. Thanks for the thorough plan/rollback design; the name-clash guard and the rewrite gate both held on a real library.

PR #454, opened 2026-08-25
feat(sync): Last.fm as a sync source with recommended playlist (#255)

### LINKED ISSUE #255: Feature: Treat Last.fm as sync source in addition to scrobble destination
Last.fm allows creating playlists and provides excellent recommendation lists. It would be nice to have last.fm show up in Sync under Sources. Specifically I would love a playlist of last.fm recommended tracks.

chaudhary-lakshay:
Closes #255.

Last.fm was scrobble-only. This makes it a **source**: connecting it surfaces a Last.fm card under Sync → Sources, and its recommendations build into a "Recommended by Last.fm" playlist that shows up with the other mixes.

The recipe is managed by `LastFmRecommendationSource`. Deactivating never deletes — the recipe keeps its `playlistId` and discovery backlog and the playlist keeps its rows, so disconnecting hides the source and reconnecting restores it in place. `reconcile()` is idempotent and re-derives state from session + toggle + the Stash-Mixes master switch.

No schema change: this branch adds DAO queries only, and the database stays at `version = 42`.

## Review fixes

Two defects found in review, both fixed here:

- **Bypassed the Stash Mixes master opt-out.** The recipe materializes a `STASH_MIX` but is not builtin, and the opt-out only sweeps builtins (`setActiveForBuiltins` / `setActiveForBuiltinMixes`), so connecting Last.fm lit the mix up even with mixes switched off. Activation is now gated inside `applyActivation` — the single place activation is applied, so both entry points inherit it. Covered by a new test with a negative-control run.
- **Re-enabling mixes left it hidden, and connecting left it empty.** Turning the master switch back on only reactivated builtins, and finishing auth mid-session reconciled with `kickRefresh = false` so the playlist sat empty until a restart. Both master-switch branches now call `reconcile()` — one predicate rather than a second copy to drift — and the interactive connect passes `kickRefresh = true`.

## Not verified end-to-end

**This has never been exercised on a device.** The build gates the whole feature on a Last.fm API key: `app/build.gradle.kts` reads `lastfm.apiKey` / `lastfm.apiSecret` from `local.properties` (or `LASTFM_API_KEY` / `LASTFM_API_SECRET`) into a `buildConfigField`, and without one the app correctly reports *"This build of Stash doesn't include a Last.fm API key."*

The negative case does check out — no Last.fm card appears under Sources while disconnected. Everything past connecting needs a key plus a Last.fm account, so **the connect → recommendations → playlist path is unverified outside unit tests.**

## Known gap

`ensureRecipe()` claims the managed row by display name, so a user's own Stash Mix named exactly "Recommended by Last.fm" would be adopted and toggled by connect/disconnect. A stable marker column would fix it properly; that's a schema change and isn't in this branch.

## Tests

`core:data` 720, `feature:sync` 45, `feature:settings` 38 — 0 failures; `:app:compileDebugKotlin` clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ These sources are somebody else's infrastructure, mostly run solo and mostly fre
 - **Spotify + YouTube Music in one unified library** — liked songs, playlists, daily mixes, every Spotify mix worth syncing
 - **Bulletproof matching** — finds the right version of a track 99% of the time
 - **Last.fm scrobbling**, optional, off by default
+- **Last.fm recommendations** — connect Last.fm and it appears under Sync → Sources, keeping a rotating "Recommended by Last.fm" playlist of tracks it suggests from your listening
 - **Wrong-match flag** — if Stash picked the wrong version, tap once from Now Playing and it queues a re-search
 - **Likes and History mirroring** — When enabled, each track you like & stream in Stash lands in your Spotify & YouTube accounts.
 
--- a/app/src/main/kotlin/com/stash/app/StashApplication.kt
+++ b/app/src/main/kotlin/com/stash/app/StashApplication.kt
@@ -118,6 +118,14 @@ class StashApplication : Application(), Configuration.Provider {
     @Inject
     lateinit var stashMixPreference: com.stash.core.data.prefs.StashMixPreference
 
+    /**
+     * Issue #255 — Last.fm-as-a-source lifecycle owner. Startup reconcile
+     * activates/deactivates the "Recommended by Last.fm" recipe to match
+     * the current session + preference state.
+     */
+    @Inject
+    lateinit var lastFmRecommendationSource: com.stash.core.data.mix.LastFmRecommendationSource
+
     @Inject
     lateinit var downloadNetworkPreference: DownloadNetworkPreference
 
@@ -395,6 +403,12 @@ class StashApplication : Application(), Configuration.Provider {
             maybeRetuneStashMixes()
             maybeRemoveRetiredBuiltinMixes()
             maybeCleanupDiscoveryLibraryHits()
+            // Issue #255: activate/deactivate the "Recommended by Last.fm"
+            // recipe BEFORE the one-shot refresh below, so a freshly
+            // reconciled recipe is picked up by that same run instead of
+            // waiting for tomorrow's periodic cycle.
+            runCatching { lastFmRecommendationSource.reconcile() }
+                .onFailure { Log.w("StashStartup", "lastfm recommendation reconcile failed", it) }
             // Fire a one-shot refresh on first launch so mixes populate
             // without waiting for the 24-hour periodic cycle. Subsequent
             // one-shots are safe (unique-work policy = REPLACE).
--- a/core/data/src/main/kotlin/com/stash/core/data/db/dao/PlaylistDao.kt
+++ b/core/data/src/main/kotlin/com/stash/core/data/db/dao/PlaylistDao.kt
@@ -350,6 +350,10 @@ interface PlaylistDao {
     @Query("SELECT * FROM playlists WHERE id = :id LIMIT 1")
     suspend fun getById(id: Long): PlaylistEntity?
 
+    /** Reactive twin of [getById] — emits again whenever the row changes. */
+    @Query("SELECT * FROM playlists WHERE id = :id LIMIT 1")
+    fun getByIdFlow(id: Long): Flow<PlaylistEntity?>
+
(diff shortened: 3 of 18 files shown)


rawnaldclark:
Now conflicts with master in `SettingsViewModel.kt` and `SettingsViewModelTest.kt` (the lossless-settings changes that merged this week). Please rebase onto master; it also carries #447's commits, so rebase after #447 merges. The unverified connect → recommendations → playlist path can be tested on my side — I have a Last.fm key locally — so once it rebases cleanly I'll take it through on a device.

rawnaldclark:
Same stacking note as on #452: this branch still carries #447's two commits (`380177c4`, `0675bf53`), which went into master as a squash, so a plain rebase replays them into conflicts. Use:

```
git fetch origin master
git rebase --onto origin/master 0675bf53
```

Master also gained a schema migration to version 43 today (`MIGRATION_42_43`, data-only). If this PR adds any migration of its own, it becomes 43→44 with a regenerated `44.json`.

chaudhary-lakshay:
Nothing to action here beyond the rebase, and that's the one thing I'm currently blocked on — a local tooling permission, not a conflict. I have your recipe and it's queued behind #452.

Two things confirmed while working the other two PRs, both matching your notes:

- #447 landed as a merge commit (`0357b653`), so `380177c4` and `0675bf53` are already ancestors of master. `git rebase --onto origin/master 0675bf53` and a plain rebase are equivalent here — using your form regardless, since it's explicit about replaying only this branch's own work.
- Master is at `version = 43` with `MIGRATION_42_43` as the tail. This branch adds no migration of its own, so there's nothing to renumber; if that changes it becomes 43→44 with a regenerated `44.json`, as you said.

Once it's up, the connect → recommendations → playlist path is yours to take through on a device — I have no Last.fm key here, so that stays unverified on my side.

chaudhary-lakshay:
Rebased and pushed. Two conflicts, both in the settings layer and both additive rather than either/or: master added `relayClient` to `SettingsViewModel`'s constructor while this branch added `lastFmRecommendationSource`, and the same pair in `SettingsViewModelTest`'s `newVm()`. Resolved as a union — dropping either would have broken the other feature.

`:app:assembleDebug`, `:feature:settings:testDebugUnitTest` and `:core:data:testDebugUnitTest` all green after the rebase. No migration added here, so master's 43 stands.

The connect → recommendations → playlist path on a device is yours whenever you're ready.

rawnaldclark (commented):
Read the source end to end: `LastFmRecommendationSource`, the recipe activation and the Stash-Mixes master-switch interplay, the worker priority, the repository reconcile hooks, and the Sync/Settings wiring. The shape is right, the two review fixes you describe are in, and "deactivating never deletes" is the correct contract for a source that can come back.

Not merging tonight, for the reason you flagged yourself: the connect → recommendations → playlist path has not run on a device. A source that builds a playlist from a third-party API needs one real pass before it ships to everyone. I'll run it on the debug phone with a key and a Last.fm session and merge it for the release after v0.9.104; if you get to a device run first, post the log of the first refresh (recipe id, track count, and the "Recommended by Last.fm" playlist appearing under Sources) and I'll merge on that.

One request for when you next touch it: the known gap you note (adopting a user's own mix named "Recommended by Last.fm") is cheap to close now with a marker column or a reserved source-id on the recipe, and it will be harder once rows exist in the wild.

rawnaldclark:
Merged, via a local merge onto today's master. Device pass on the Pixel 6 (debug build of master + this branch):

- Connect → Allow on last.fm → Finish connecting → `auth.getSession` 200, session saved, `LastFmRecSource: reconcile: connected=true enabled=true`, `StashMixRefreshWorker` kicked at once.
- First refresh built "Recommended by Last.fm" as a 40-track STASH_MIX (73 discovery rows, 0 failed) on an account with zero scrobbles, so the seeding from the library holds up.
- Sync tab shows the source card, "Recommended tracks", the explainer, and "Manage Last.fm playlists".

Two things you'll want to know:

1. **Conflicts on rebase.** Master changed `ManagePlaylistsScreen` under you last night (15ffe672: a `pinnedToHomeAt` column on `ManageRow`, and a `homeVisibilityLabel` helper at the file's tail). I resolved both as a union: your three-way source branch keeps master's seventh `ManageRow` argument, and both tail additions stay. Nothing of yours was dropped.
2. **The card appeared only after a cold restart** in my run; the Sync tab I had open during the connect didn't pick the new session up live. Not blocking (the state does update on the next screen creation), but worth a look at `observeState()`'s subscription if you touch this again, along with the marker column for the recipe.

Thanks for the source; it's a good one.

PR #458, opened 2026-08-30
Fix: Pin debug signing to a shared keystore

ParaliyzedEvo:
## Summary
Debug builds were signing with each machine's local
`~/.android/debug.keystore`, which is per-machine and not checked into the
repo. Android enforces one signing certificate per `applicationId`
device-wide (shared across all user profiles — main, Secure Folder, work
profile, etc.), so installing a debug build signed with a different
keystore than whatever's already on the device anywhere fails with
`INSTALL_FAILED_UPDATE_INCOMPATIBLE` (`result of install: -7`). This
surfaced as debug installs failing into Secure Folder while an
out-of-sync copy remained on the main profile — nothing to do with Knox
policy, just a plain cert mismatch on the shared package record. Release
builds never hit this since `hasReleaseSigning` already pins a single
checked-in/CI keystore.

## Changes
* `app/build.gradle.kts`
  * Added a `debug` entry to `signingConfigs` pointing at a checked-in
    `debug.keystore` (standard AOSP debug cert: alias `androiddebugkey`,
    store/key password `android`).
  * Wired `buildTypes.debug.signingConfig` to the new config.
* `debug.keystore` (repo root, alongside `keystore.properties`)
  * New shared debug keystore, generated via `keytool` with the standard
    AOSP debug-cert parameters. Intentionally not gitignored — same
    convention as the default per-machine debug key, just shared instead
    of local.
* `.gitignore`
  * Added `!debug.keystore` exception to the existing `*.keystore` rule.

## Test plan
* [x] Confirm `assembleRelease` is unaffected (still uses
      `keystore.properties`/CI env vars, falls back to debug cert only
      when release signing is unconfigured).


rawnaldclark (changes requested):
Thanks for this — the diagnosis is right and the mechanism is the right one: one committed debug key so every machine and CI sign debug builds identically, no more `INSTALL_FAILED_UPDATE_INCOMPATIBLE` between builders.

## The blocker: the key itself

`debug.keystore` in this PR was generated fresh on 2026-08-29 (`keytool -list`: valid from Aug 29 2026, SHA-256 `80:0F:72:0A:…`). The alias, DN and password follow the AOSP convention, but the key material is new — there is no "standard AOSP debug cert" file, only standard parameters. So the first build from this branch is a stranger to **every existing debug install**: mine, yours before you switched, the Pixel 5 test rig, and every Discord tester who sideloaded a debug artifact. Each one hits the exact error this PR is meant to fix, once, and the only way through is uninstall.

Uninstall wipes app data, and for debug installs that is not just accounts: downloads live in app-internal storage by default (`FileOrganizer`: `context.filesDir/music`), and the in-app backup deliberately excludes audio. My daily debug install holds 861 tracks / 32.9 GB. That cannot ship as a silent side effect.

## Fix

Swap the file for the maintainer's existing `~/.android/debug.keystore` (I'll push it to this branch, or send it to you — your call). Every device already updated from this PC keeps working; your devices and CI reinstall once, which this PR already accepts. Same alias/password, so the Gradle block is unchanged.

Two smaller things:

- The PR body's "Appended `MIGRATION_42_43` to `ALL_MIGRATIONS`" line belongs to another PR — nothing in this diff touches `StashDatabase.kt`.
- Add one paragraph to CONTRIBUTING saying debug builds are signed with the committed key and that a debug install from before it needs a one-time reinstall. Committing a debug key to a public repo is the normal trade-off (it is a debug key with the well-known password, same as every local one); worth a sentence so nobody mistakes it for a leak.

Release signing is untouched by this PR, so shipped users are unaffected either way — the constraint above is about debug installs only.

ParaliyzedEvo:
@rawnaldclark Added the note in the md doc. Push your ~/.android/debug.keystore onto the pr branch and replace the current debug.keystore at repo root

Issue #16, opened 2026-04-23
New feature idea


juv3nal:
Since all the tracks are saved locally, would it be possible to add a crossfade effect when changing tracks?
Amazing app, thank you!

rawnaldclark:
Will be added soon! Plus I've noticed some tracks had a dead space ~2-3 seconds before playing, I'll eliminate that too.

Thanks for the feedback!!

juv3nal:
Excellent. Most streaming apps do it only when the tracks transition from one to another, however it would be amazing if this effect can be applied also when tracks are manually changed. Not sure if this possible. Many thanks.

juv3nal:
Hi, just wondering if this is still in the works. There's so much going on with the app that you might have forgotten about it. 😅
You're doing an fantastic job, many thanks.

rawnaldclark:
i honestly did forget.. i thought i shipped the crossfade already, however that was volume normalization not the crossfader. my bad! and i appreciate that :)

should be ready by the end of the week.

juv3nal:
Please remember the cross-fade feature. You're doing an amazing work. Thank you.

Issue #15, opened 2026-04-20
Playback Bug


TylorRKO:
Another issue I’ve been running into is that a large number of songs do not automatically start when the previous track ends. It moves to the next song, but playback gets stuck there and does not begin.

When this happens, the play button usually does not respond either.

The only workaround I’ve found is to manually skip to another track, or skip forward and then back to get playback working again.

Not sure if this is a queue/playback bug, but I wanted to report it since it happens fairly often.

Thank you!

PR #472, opened 2026-09-07
Add discord rpc

ParaliyzedEvo:
## Summary

Adds optional Discord Rich Presence to Stash, allowing users to connect their own Discord account and have Stash show what they're currently playing as their Discord status.

The integration uses Discord's OAuth flow with `sdk.social_layer_presence` and the Discord headless-session RPC approach. No Discord bot or bot token is used. The previous `rpc.activities.write` scope was found to be Discord-locked for third-party apps and caused `invalid_scope` regardless of PKCE configuration, so the integration was switched to `sdk.social_layer_presence`, which requires enabling Discord Social SDK for the application in the Developer Portal.

The RPC implementation is based on the approach used by [[echo-discord](https://github.com/brahmkshatriya/echo-discord)](https://github.com/brahmkshatriya/echo-discord), with upstream issues [[#15](https://github.com/brahmkshatriya/echo-discord/issues/15)](https://github.com/brahmkshatriya/echo-discord/issues/15) (rate-limit/debounce handling) and [[#16](https://github.com/brahmkshatriya/echo-discord/issues/16)](https://github.com/brahmkshatriya/echo-discord/issues/16) (incomplete API headers and v10 endpoints) addressed as part of the port.

`core:data` already depends on `core:auth` for `TokenManager`, so Discord header construction, rate limiting, token validation, and OAuth/auth state live in `core:auth/discord`. The ongoing presence client and playback integration live in `core:data/discord`.

The integration also handles several non-obvious Discord/WebView/RPC edge cases:

* Discord's web client is a React SPA. After login it changes from `/login` to `/channels/@me` using `history.pushState()`, so `WebViewClient.onPageFinished` does not fire. Token extraction instead hooks `doUpdateVisitedHistory`, which Chromium's WebView invokes for the route change.
* Discord's account switcher breaks the classic flat `localStorage.getItem("token")` lookup when multiple accounts are present. Extraction falls back to Discord's in-memory `TokenManager` module via its webpack module graph, filtered on `exports.setToken` to avoid matching unrelated modules.
* Discord's `token` localStorage value is JSON-stringified, while `evaluateJavascript` adds another JSON encoding layer. Both layers need to be unwrapped or the extracted token contains literal escaped quotes and silently fails validation.
* `headless-sessions` requires both a registered `application_id` and a `platform` field on the activity object. Omitting either produces a generic `400 Invalid Form Body`.
* The OAuth handshake is triggered proactively when Discord is connected rather than waiting for the first playback update, so the one-time consent prompt appears during account linking.
* Rapid playback/seek updates are debounced and rate-limited so Discord is not spammed during track skipping or scrubbing.

### Authentication / OAuth

* `core/auth/model/AuthModels.kt`

  * Added `DISCORD` to `AuthService`.

* `core/auth/TokenManager.kt`

  * Added `discordAuthState`, `getDiscordUserToken`, and `connectDiscordWithToken`.

* `core/auth/TokenManagerImpl.kt`

  * Added Discord auth state observation.
  * Added connect/clear/`isAuthenticated` wiring.
  * Added profile validation with failure logging.

* `core/auth/store/EncryptedTokenStore.kt`

  * Added encrypted `discordToken` / `discordUser` flows.
  * Added `saveDiscordToken` / `clearDiscord`, following the existing Spotify/YouTube storage pattern.

* `core/auth/discord/DiscordHeaders.kt`

  * Added a realistic Discord client header set including User-Agent, `X-Super-Properties`, locale/timezone, etc.
  * Fixes the incomplete-header issue from echo-discord #16.

* `core/auth/discord/DiscordRateLimiter.kt`

  * Added a shared cooldown driven by `Retry-After`, `X-RateLimit-Remaining`, and `X-RateLimit-Reset-After`.
  * Handles the rate-limit/backoff portion of echo-discord #15.

* `core/auth/discord/DiscordProfileValidator.kt`

  * Added connect-time validation of scraped Discord tokens using the `oauth2/authorize` GET flow rather than directly hitting `/users/@me`.

### Discord RPC / Presence

* `core/data/discord/DiscordModels.kt`

  * Added `DiscordActivity`, `DiscordSession`, and `DiscordActivityType` wire models.

* `DiscordRpcConfig.kt`

  * Changed OAuth scope from the Discord-locked `rpc.activities.write` to `sdk.social_layer_presence`.

* `core/data/discord/DiscordRpcClient.kt`

  * Added OAuth `connect()` which proactively calls `getAccessToken()` so the OAuth handshake and one-time consent prompt happen when Discord is linked.
  * Added presence posting/clearing through `headless-sessions`.
  * Uses v9 endpoints throughout.
  * Added debouncing for rapid `requestActivity` calls, collapsing fast track skips/scrubbing into a single network request for the latest state.
  * Applies the shared Discord rate limiter.

* `core/data/discord/DiscordRpcCoordinator.kt`

  * Added singleton ownership of the active `DiscordRpcClient`.
  * Observes `discordAuthState`.
  * Exposes `updateNowPlaying` for playback.
  * Extended `updateNowPlaying` with `album`, `positionMs`, and `durationMs`.
  * Sends the album as `assets.largeText` instead of duplicating the artist.
  * Omits `largeText` when the album is blank.
  * Adds Discord timestamps from playback position/duration to provide a seekbar.
  * Only sends timestamps when `durationMs > 0` to avoid bogus bars for an unprepared player.

### Playback integration

* `core/media/.../service/StashPlaybackService.kt`

  * Hooks `updateDiscordPresence` into:

    * `onMediaItemTransition`
    * `onIsPlayingChanged`
    * crossfade swap callback
    * `onPositionDiscontinuity` for user-initiated seeking
  * Reads album title, current position, and duration from the master player.
  * Seek events are still protected by the existing RPC debounce to avoid network spam while dragging the seekbar.

### Settings / UI

* `feature/settings/components/DiscordLoginWebView.kt`

  * Added Discord WebView login matching the existing Spotify login flow.
  * Uses `doUpdateVisitedHistory` for SPA-safe token extraction.
  * Handles Discord's multi-account switcher through the webpack `TokenManager` fallback.
  * Safely unwraps the double JSON encoding produced by `localStorage` + `evaluateJavascript`.

* `feature/settings/components/DiscordTokenDialog.kt`

  * Added manual token-paste fallback when WebView login is unavailable.

* `feature/settings/SettingsViewModel.kt`

  * Added Discord connect/disconnect actions and local dialog state.
  * Wired Discord state into the combined `uiState`.
  * Updated `onDisconnectDiscord()` for the expanded `updateNowPlaying` signature, passing blank/zero playback values so disconnect safely clears presence.

* `feature/settings/SettingsUiState.kt`

  * Added `discordAuthState` and WebView/token-dialog state.

* `feature/settings/SettingsAccountsScreen.kt`

  * Added Discord `AccountConnectionCard`.
  * Added WebView and token-dialog overlays.

* `feature/settings/src/test/.../SettingsViewModelTest.kt`

  * Updated tests for the new `discordRpcCoordinator` constructor dependency.

* `app/src/main/kotlin/com/stash/app/StashApplication.kt`

  * Injects and starts `DiscordRpcCoordinator`, following the existing `lastFmScrobbler.start()` pattern.

## Test plan

* [x] WebView login works with both fresh sessions and Discord's multi-account switcher.
* [x] Manual token-paste fallback works.
* [x] OAuth authorization/consent completes successfully with `sdk.social_layer_presence`.
* [x] Settings correctly reflects Discord connection/disconnection state.
* [x] Presence shows song, artist, album, and an accurate seekbar.
* [x] Track changes, pause, seeking, and disconnect correctly update/clear presence.
* [x] Rapid skipping/seeking is debounced and does not trigger 429s.
* [x] Device / Android version tested: S26 Ultra / Android 16.


rawnaldclark:
Merged as `5cdb3038`, with one addition.

Connecting now asks first. Discord has no sanctioned way for a phone app to set Rich Presence — this PR says so plainly, which is why the account's own session is used — and Discord's [Platform Manipulation Policy](https://discord.com/safety/platform-manipulation-policy-explainer) names that pattern a self-bot ("user accounts modified to perform automated actions"), while their [support article](https://support.discord.com/hc/en-us/articles/115002192352-Automated-User-Accounts-Self-Bots) says it can end in account termination. I checked both pages rather than going from memory.

Enforcement in practice goes after spam, not status updates, so the real-world risk is low, and the dialog says exactly that. But the account at stake is the user's, so the choice is theirs to make knowingly: a dialog naming the rule, calling the risk unlikely, and noting that disconnecting clears the status and deletes the token. Cancel is the quiet default; the confirm reads "Connect anyway". The README gained the same warning above your existing explanation, with links to both pages.

One merge conflict, resolved as a union: master and this branch each added a constructor argument to the same `SettingsViewModelTest` construction.

Verified on a Pixel 6: the app launches clean on the merged tree, the dialog appears on Connect, and Cancel closes it without connecting. Suites green in `:feature:settings`, `:core:data` and `:core:auth`.

Thanks for the care in the write-up — the rate limiter, the backoff on the headless-sessions 429s, and being upfront about the token in your own README made this straightforward to review.

Issue #43, opened 2026-05-12
Not seeing all YT music playlists


LucasMiserez:
I've added my YT music cookie, but I only see my "loved tracks" playlist, I have 3/4 other playlists but they are not visible anywhere in the app. 

I've look through the documentation but could also not find anything. Is this a bug or something that I am doing wrong?

rawnaldclark:
let me know if this update fixed your issue.

LucasMiserez:
> let me know if this update fixed your issue.

I've installed the latest version, but I still have the same issue where it only shows the "New Release Mix" as an option

rawnaldclark:
Closing per maintainer triage. If you're on the latest version and still only see a subset of your YouTube Music playlists, please open a fresh report with your app version and a screenshot — thanks! 🙏

PR #486, opened 2026-09-20
fix: YouTube saved-albums library only synced the first page

### LINKED ISSUE #276: App only syncs one page of yt albums
https://github.com/rawnaldclark/Stash/issues/113#issuecomment-5628665709

---
_Imported from Discord: https://discord.com/channels/1508402965042561054/1547801638968762428_

ParaliyzedEvo:
## Summary
`YTMusicApiClient.getSavedAlbums()` fetches the Library → Albums tab
(`FEmusic_liked_albums`), which InnerTube renders as a `gridRenderer`.
`paginateBrowse()`'s continuation walk is driven entirely by
`extractContinuationToken()`, which only knows how to read a token out of
shelf-shaped responses (`musicShelfContinuation`,
`musicPlaylistShelfContinuation`, `appendContinuationItemsAction`,
`musicShelfRenderer`) — it has no `gridRenderer`/`gridContinuation` case.
So on a grid response `token` comes back null on page 1, the `while` loop
in `paginateBrowse` never runs, and any library past the first page (~100
albums) was silently dropped. `parseSavedAlbumsContinuationPage()` already
existed to parse a grid continuation page, but was dead code — nothing
could ever produce a token to reach it.

A previous attempt at surfacing this (`gridHasUnhandledContinuation()`)
detected the trailing `continuationItemRenderer` on the initial response
and flagged the fetch `partial = true` as a workaround, but that renderer
is present on page 1 of every multi-page grid — including once pagination
actually works — so it would report `partial` forever. That in turn feeds
`DiffWorker.snapshotUnreliable()`, which skips stamping `snapshot_id` and
track count on a partial snapshot, so the albums library would never
settle and would re-diff from scratch on every sync.

Also found: `getUserPlaylists()` (`FEmusic_liked_playlists`) hits the
identical gap when the web client renders it as a grid instead of a
`musicShelfRenderer`, and `parseSavedAlbumsContinuationPage` only read the
legacy `continuationContents.gridContinuation` envelope — current
InnerTube responses can serve a grid continuation as
`onResponseReceivedActions[*].appendContinuationItemsAction`, which would
still parse to zero albums even once a token is found.

## Changes
* `YTMusicApiClient.kt`
  * `extractContinuationToken()`: added grid-shape handling — a
    `continuationContents.gridContinuation` envelope (shelf-level
    `continuations` or a trailing `continuationItemRenderer` item), and an
    initial singleColumn browse whose section is a `gridRenderer` (bare or
    wrapped in `itemSectionRenderer`).
  * `getSavedAlbums()`: removed the `gridHasUnhandledContinuation` partial
    override; `paginated.partial`/`partialReason` from `paginateBrowse` is
    now the sole signal, same as every other fetch path. Added
    `.distinctBy { it.id }` on the accumulated items — grids repeat entries
    across continuation boundaries and the per-page parser only dedupes
    within a page.
  * Deleted `gridHasUnhandledContinuation()` — dead now that real grid
    pagination works, and actively wrong (see Summary) if left in.
  * `parseSavedAlbumsContinuationPage()`: now reads both the legacy
    `continuationContents.gridContinuation.items` envelope and
    `onResponseReceivedActions[*].appendContinuationItemsAction.continuationItems`.
* `PlaylistFetchWorker.kt`
  * `fetchAndSnapshotSavedAlbums()`: updated the comment referencing the
    now-deleted `gridHasUnhandledContinuation`; behavior unchanged —
    `markYoutubeIncomplete` still gates on `paged.partial`.

## Follow-ups (not in this PR)
* `parseUserPlaylistsContinuationPage()` has the same single-envelope gap
  as the old `parseSavedAlbumsContinuationPage` and is now reachable for a
  grid-shaped Playlists tab via the `extractContinuationToken` fix above —
  needs the same append-action handling.
* `InnerTubeClient.browseWithStatus()` sends the continuation token only as
  a query param (`?ctoken=...&continuation=...`). This matches shelf
  continuations in production but is unverified against a real grid
  continuation response — if page 2 comes back empty/400 in testing, also
  send `continuation` in the POST body.

## Test plan
* [x] Account with >100 saved albums: confirm all pages are fetched (check
      `getSavedAlbums` album count against YT Music's own library count).
* [x] Confirm `sync_history.diagnostics` shows `getSavedAlbums` as
      `SUCCESS` with `errorMessage = null` on a healthy multi-page run (no
      longer permanently partial).
* [x] Confirm a second sync run reuses/updates the same album playlist rows
      rather than re-creating them (validates `snapshot_id`/track count are
      now being stamped).
* [x] Small-library account (<100 albums, single page): confirm no
      regression — album count and content unchanged.

Closes https://github.com/ParaliyzedEvo/Stash/issues/276


rawnaldclark (changes requested):
Nice find: the grid had no continuation case, and teaching `extractContinuationToken` about `gridContinuation` and the trailing `continuationItemRenderer` is the right fix. Two things before it merges:

**1. Playlists now page too, but their pages parse to nothing.** `extractContinuationToken` is shared by `getUserPlaylists`, `getSavedAlbums` and `getPlaylistTracks`. The library playlists page (`FEmusic_liked_playlists`) is also a grid, so with shape 5b `getUserPlaylists` now gets a token and fetches page 2+. But `parseUserPlaylistsContinuationPage` only reads `continuationContents.musicShelfContinuation`, so every extra page yields zero playlists: up to `MAX_PAGES` wasted requests per sync, and still only page 1's playlists. Could you give it the same two envelopes you added to `parseSavedAlbumsContinuationPage` (`gridContinuation.items` and `onResponseReceivedActions[].appendContinuationItemsAction.continuationItems`)? That would very likely also fix the long-standing "YouTube only syncs ~25 playlists" report (#43/#21), and the same `distinctBy` guard would apply there too.

**2. Tests.** This is shared parsing on the sync path, and `YTMusicApiClientTest` already has `extractContinuationTokenForTest` cases. Please add fixtures for:
- a singleColumn initial response with a `gridRenderer` (bare, and nested in `itemSectionRenderer`) whose last item is a `continuationItemRenderer`, returning the token;
- a `gridContinuation` page, returning its token;
- an `appendContinuationItemsAction` grid page, parsed to albums (and playlists, after 1).

Nit: the comment block above `val albums = paginated.items.distinctBy { it.id }` is indented one level too deep on its first line.

ParaliyzedEvo:
@rawnaldclark did the changes you said, lmk if there are anything else before this pr needs to be merged

rawnaldclark (approved):
Thanks, this covers everything: playlist pages in all three envelopes, dedupe across pages, and good fixture coverage for every token shape. This likely fixes the ~25-playlist YouTube sync cap (#43/#21) too.

Issue #14, opened 2026-04-20
App crashes when deleting songs in library tab


walsoni:
Hello, the first thing i wanted to say is that i searched for that kind of app forever and i'm glad that finally could find this gem.

The problem i have is when i'm deleting playlist in the library tab using option highlighted in red (delete playlist and songs) the app crashes after i accept. I dont know if that deletes the songs, but the playlist is removed from the library tab after i open stash again.

Everything works when deleting or deleting and blocking from Home tab.

Issue #13, opened 2026-04-20
Playlist Sorting


sangakuu:
Just like Spotify, add a way to sort Playlists from date added (newest to oldest), artists, alphabetical etc.

on a sidenote
awesome app you created THANK U ALOT

Issue #11, opened 2026-04-18
Use different file type


thealonealex:
WEBM is a weird file type to use, as many music players don't support it and straight up don't show the song files as options.

I propose to use something more conventiona like mp3 or m4a

rawnaldclark:
when downloading on max quality, AAC 256kbp, M4A files are downloaded when available. This should be roughly 90% of the time.

Issue #10, opened 2026-04-18
Sync preferences doesn't match


mastercpo:
<img width="317" height="529" alt="Image" src="https://github.com/user-attachments/assets/355d9399-dffe-4cb4-807b-8ef81dd44c07" />

hi, I only want to download tracks in my liked songs but the track count doesn't seem to be correct and I'm guessing it also downloads the other tracks in playlists, those are not selected I remove them individually, and they reappear again so it was not a temporary fix either

rawnaldclark:
Fixed in v0.3.7: https://github.com/rawnaldclark/Stash/releases/tag/v0.3.7

You were right — the app was downloading everything by default and unchecking didn't actually prevent future re-downloads the way you'd expect. Two things changed:

1. **Every new YouTube playlist is now disabled by default.** First tap of Sync Now is a discovery pass: it finds your mixes, Liked Songs, and library playlists (also new in this release — see #9) but downloads nothing. The Sync Preferences card auto-opens in amber with "Found N playlists — enable the ones you want, then tap Sync Now again."
2. **Tapping Sync Now a second time** only downloads the playlists you enabled. This matches how Spotify has always worked in the app.

One caveat, in full honesty: if you already have unwanted tracks on disk from earlier syncs, this release doesn't clean them up for you — your existing `sync_enabled` flags are preserved so the app doesn't destructively remove anything behind your back. A "Clean up tracks from disabled playlists" action is the next piece I want to design properly (bulk-delete needs a careful UX). For now, disabling a playlist stops future downloads; removing existing files is a manual step.

Thanks for the clear repro and for flagging the UX break.

rawnaldclark:
Quick follow-up on the cleanup part — I just re-checked the code and there's already a way to remove unwanted tracks from disk, I'd just forgotten about it when I wrote the first comment.

**On the Home screen, long-press any playlist card** → you'll get a "Delete" option. Confirming it deletes every downloaded song in that playlist from your device (files + library rows), then removes the playlist itself. That's the cleanup path for your earlier over-downloads.

So between v0.3.7's opt-in default (stops the problem going forward) and the existing long-press delete (cleans up what's already there), there's nothing more I need to ship. Sorry for the confusion in the earlier comment.

rawnaldclark:
v0.3.8 is up with the proper cleanup story: https://github.com/rawnaldclark/Stash/releases/tag/v0.3.8

Two pieces directly for the "unwanted download" problem:

[...]

PR #489, opened 2026-09-20
Security Fixes

ParaliyzedEvo:
## Summary
All five commits are Copilot Autofix responses to "broken/risky cryptographic algorithm" scanning alerts. Four touch `JioSaavnClient.kt`'s `decrypt320Url()`/`decryptTemplate()` (alerts #9, #10, #14, #13, applied in that order — the diffs chain cleanly: 9→10→14→13 by blob hash) and one touches `qobuz.js`'s signing helper (#11). Two more are routine dependabot bumps.

**⚠️ Likely functional regression, not just hardening.** `decrypt320Url()`'s own comment (added in #9) says JioSaavn's `encryptedMediaUrl` "currently uses legacy DES/ECB format… retained strictly for protocol compatibility with upstream payloads." The alert chain then:
- **#10** adds an AES/ECB attempt (key = `DES_KEY` truncated to 16 bytes) *ahead of* the DES/ECB path, keeping DES as a fallback.
- **#14** deletes the DES fallback, leaving AES/ECB as the only path.
- **#13** changes that sole path from AES/ECB/PKCS5Padding to AES/GCM/NoPadding, treating the first 12 bytes of the payload as a nonce.

None of this changes what JioSaavn actually sends. If the upstream payload is still DES/ECB-encrypted (as #9's own comment says), decryption now fails for every real request — Copilot Autofix "fixed" the algorithm without re-deriving it from the actual protocol, and #14 threw away the working fallback. Same shape of risk in `qobuz.js`: `md5` → `sha256` for `request_sig` assumes Qobuz's API will accept a different signature algorithm than the one its own protocol defines; if Qobuz's servers expect MD5-based signatures, this breaks `getFileUrl` auth regardless of local correctness. Worth manually verifying both against live services before merging.

**Dependency bumps** (dependabot, mechanical): `sharp` 0.35.2 → 0.35.4, `wrangler` 4.120.0 → 4.135.0, applied identically in `infra/tipjar-worker` and `infra/lastfm-proxy`.

## Changes
* `JioSaavnClient.kt`
  * **#9**: explicit `Charsets.UTF_8` on `DES_KEY.toByteArray()` (was platform-default charset); `trim()` + explicit `runCatching` Base64 decode with null-safe bailout; `@Suppress("DEPRECATION")` + comment documenting DES/ECB as required for upstream compatibility.
  * **#10**: extracts `decryptTemplate(payload, key, algorithm, transformation)`; tries AES/ECB/PKCS5Padding (16-byte-truncated key) first, falls back to DES/ECB/PKCS5Padding.
  * **#14**: removes the DES/ECB fallback branch added in #10 — AES/ECB is now the only path.
  * **#13**: drops the now-unused `algorithm`/`transformation` params (hardcodes AES); switches `decryptTemplate` to AES/GCM/NoPadding, splitting the payload into a 12-byte IV prefix + ciphertext+tag. Adds an unused `kotlin.math.max` import (dead code, never referenced in the diff).
* `infra/lossless-relay/src/qobuz.js`
  * **#11**: `md5()` helper now calls `createHash("sha256")` internally but keeps the `md5` name — misleading given callers still refer to it as `md5(s)`.
* `infra/tipjar-worker/package.json`, `package-lock.json` — `sharp` → 0.35.4, `wrangler` → 4.135.0.
* `infra/lastfm-proxy/package.json`, `package-lock.json` — `sharp` → 0.35.4, `wrangler` → 4.135.0.


rawnaldclark (changes requested):
Thanks for pulling these together and for flagging the crypto risk in the description. You were right to: I tested it, and one change has to come out before this can merge.

**Blocking: `infra/lossless-relay/src/qobuz.js` (alert #11).** `md5()` now hashes with SHA-256, but `request_sig` for `track/getFileUrl` is defined by Qobuz's protocol as MD5, and Qobuz's servers verify exactly that. Running the relay's own suite against this branch (`cd infra/lossless-relay && npm test`), the `signGetFileUrl reproduces the client's HAR vectors` test fails: it expects `013c10042c5e15ca5f1d85610bdd62ad` (MD5, captured from the real Qobuz app) and gets a 64-hex SHA-256. Deployed, every mint would be rejected and lossless would stop for everyone. CI passed only because no CI job runs the relay's node tests. Please revert this line.

**JioSaavn (alerts #9/#10/#14/#13).** The final state is DES/ECB again after your revert commit, which is correct: the algorithm is set by JioSaavn's payload, not by us. What's left is harmless, but please:
- drop the now-unused imports `javax.crypto.spec.GCMParameterSpec` and `kotlin.math.max`;
- update the PR description, which still describes the intermediate AES/GCM states.

**Dependency bumps:** fine.

For the code-scanning alerts on MD5 (Qobuz signing) and DES (JioSaavn), the right answer is to dismiss them as "won't fix: required by the upstream protocol" rather than autofix them. Both algorithms are dictated by the other side, and these values are request signatures and URL obfuscation, not secrets we protect.

ParaliyzedEvo:
@rawnaldclark fixed the issues, nothing should be blocking this pr now

rawnaldclark (approved):
Thanks, all fixed: signing is back to MD5 (the relay suite passes on this merged into master), the unused imports are gone, and the dependency bumps combine cleanly with master's relay changes.

